PR #50, opened 2026-06-15
Feature: LDAP Validation Endpoint

sdewhitt:
*Issue #43*

*Revision of PR #46 (cleaner commit history)*

Allow customers to conveniently validate their LDAP broker configs via /api/aws/auth/validate/ldap with more descriptive (but safe) error messages instead of blindly guessing error causes, changing the config, and restarting the broker (~30 min per restart)

This endpoint is extensible for other forms of auth (HTTP/OAuth), however this PR focuses on solely implementing LDAP.

### Core subsystem (src/aws_auth_validate_*)
  - aws_auth_validate_mgmt -- Cowboy REST handler for PUT /api/aws/auth/validate/:method; pipeline: feature toggle → admin auth → per-IP rate limit → body-size cap → JSON decode → concurrency semaphore → registry dispatch → audit log → response.
  - aws_auth_validate_registry -- method → backend dispatch with per-method enable flags and request-field filtering.
  - aws_auth_validate_backend -- behaviour (method_name/0, validate/1, allowed_fields/0).
  - aws_auth_validate_ldap -- LDAP backend: ephemeral eldap connect → optional StartTLS → simple bind → optional post-bind DN-lookup + authorization-query checks. All outcomes collapse to fixed response categories (no credential/LDAP detail leakage).
  - aws_auth_validate_ldap_query -- self-contained parser for the authz query DSL, mirroring rabbit_auth_backend_ldap_util:parse_query/1 exactly (R12 parity).
  - aws_auth_validate_rate_limiter / aws_auth_validate_semaphore -- supervised gen_servers, started by aws_sup only when the feature is enabled.
  - aws_auth_validate_arn_lock -- supervised gen_server mutex that serializes ARN resolution. Meant to prevent region-clobbering with the global AWS credential state when resolving password/cacertfile ARNs via the validation endpoint.

### Config (priv/schema/aws.schema, aws_sup)
  - aws.auth_validation.* cuttlefish settings (toggle, body size, concurrency, rate limit, timeout, user tag); workers conditionally supervised.
  - Defaults are applied in code, not declared as schema defaults, so they aren't emitted into every generated config (keeps config_schema_SUITE snippets exact).
  
### Tests
  - aws_auth_validate_tests (eunit): rate limiter, semaphore, ARN lock (value pass-through, serialization under concurrent callers, exception re-raise, survival of a crashing closure), registry, LDAP input validation, and query-parser + upstream-parity tests.
  - aws_auth_validate_mgmt_SUITE (CT): the HTTP pipeline (auth, rate limit, body cap, semaphore, dispatch, response mapping) without a live LDAP server.
  - aws.snippets: config-schema coverage for the new settings.

### Response contract
  All failures return a fixed {"error": "<category>", "message": "<constant>"} -- never raw LDAP errors, hostnames, DNs, or the resolved password.
  Categories: input_invalid / connection_failed / tls_failed / query_invalid (400), auth_failed / config_conflict / authz_unverified (422), plus pipeline-level insufficient_user_tag (401), unknown_method / method_disabled (404), rate_limited (429), capacity_exhausted (503).

### Security notes
  - Endpoint disabled by default; admin-tag gated.
  - Ephemeral connections only -- no broker/auth-backend state mutation.
  - password_arn is resolved (via the plugin's existing ARN machinery) only after all pure input validation passes, so malformed requests never trigger a secret fetch. Resolution is serialized by aws_auth_validate_arn_lock so concurrent validations cannot clobber the shared client's region mid-resolution.
  - assume_role is deliberately not accepted (it would let a request swap the broker's global AWS credentials -- confused-deputy).

### Excluded from this PR
- Integration tests to prevent divergence from the real LDAP auth backend


src/aws_auth_validate_arn_lock.erl
the-mikedavis, line 1: For a node-local lock you can use [`global:trans/4`](https://www.erlang.org/doc/apps/kernel/global.html#trans/4) - passing `[node()]` as the node list makes it behave roughly like a gen-server does here. A gen-server is also a good way of serializing executions but I think a `global` lock might be more natural here.

src/aws_auth_validate_ldap.erl
```diff
@@ -0,0 +1,489 @@
...
+
+maybe_start_tls(_Handle, false, _SslOpts, _Timeout) ->
+    ok;
+maybe_start_tls(Handle, true, SslOpts, Timeout) ->
+    eldap:start_tls(Handle, build_ssl_opts(SslOpts), Timeout).
+
+%% Translate the JSON ssl_options map into an Erlang ssl options
+%% proplist suitable for eldap. Unknown keys are ignored to keep the
+%% allowed surface narrow.
+build_ssl_opts(Map) when is_map(Map) ->
```
the-mikedavis: Rather than ignoring unknown keys we might want to return an error? A mis-typed option could make the validation result look green despite the desired config not really being tested.

src/aws_auth_validate_ldap.erl
```diff
@@ -0,0 +1,489 @@
...
+    case maps:get(Key, Body, undefined) of
+        undefined -> {ok, Acc#{AccKey => false}};
+        Bool when is_boolean(Bool) -> {ok, Acc#{AccKey => Bool}};
+        _ -> {error, input_invalid, Reason}
+    end.
+
+parse_ssl_options(Body, Acc) ->
+    case maps:get(<<"ssl_options">>, Body, undefined) of
+        undefined ->
+            {ok, Acc#{ssl_options => #{}}};
```
the-mikedavis, line 251: I think we may want to set reasonable config for ssl options, like setting `verify` by default if not set. For historical Erlang/OTP reasons the default is verify_none which is not secure.

src/aws_auth_validate_mgmt.erl
```diff
@@ -0,0 +1,267 @@
...
+    Headers = #{<<"content-type">> => <<"application/json">>},
+    Req1 = cowboy_req:reply(401, Headers, Body, ReqData),
+    {stop, Req1, Context}.
+
+%%--------------------------------------------------------------------
+%% Request helpers
+%%--------------------------------------------------------------------
+
+peer_ip(Req) ->
+    case cowboy_req:peer(Req) of
```
the-mikedavis, line 250: If the broker is behind a load balancer like an NLB how does this behave? Would we be rate limiting the NLB or the end-user?

lukebakken: To be honest I don't think rate-limiting has _any place_ in this code (maaaaybe rate-limiting LDAP queries). My $0.02

the-mikedavis: If we serialize the LDAP queries with a gen_server or lock, I'm not so worried about rate limiting

lukebakken (commented):
*This review was performed by AI under the guidance of @lukebakken.*

---

Thanks for this. It is a well-structured, security-conscious PR: the password-leak defense (R6), the pure-before-network ordering in `validate/1`, the query-parser parity strategy (R12), and the ARN-lock rationale are all clearly reasoned and documented. Comments below are mostly edge cases and coverage gaps, not structural concerns. None block the core happy path.

## Findings

**1. OPTIONS preflight returns 401 when a custom `required_user_tag` is set (medium).** A `PUT` route also accepts `OPTIONS`, and cowboy_rest runs `OPTIONS` through `is_authorized/2`. On the custom-tag branch, `rabbit_mgmt_util:is_authorized/2` short-circuits `OPTIONS` to `{true, ReqData, Context}` with `Context.user = undefined`, so `authorize_tag/3`'s `#context{user = #user{...}}` head fails to match and falls through to a 401. The default `administrator` path is unaffected (it short-circuits earlier). The practical impact depends on whether a cross-origin browser client calls this endpoint; there is no in-tree UI consumer (`web_ui/0` is `[]`). Suggested fix: special-case `OPTIONS` (or an `undefined` user) in `authorize_tag/3`.

**2. The custom-tag authorization branch is untested (medium).** No test sets `auth_validation_required_user_tag` to a non-admin value, so `authorize_tag/3`, the `insufficient_user_tag` 401, and finding 1 are all unexercised. A regression test here would also cover finding 1.

**3. Per-IP rate limiter behind a load balancer (medium, deployment-specific).** The limiter keys on `cowboy_req:peer/1` (the immediate TCP peer); the management listener reads no `X-Forwarded-For` and enables no `proxy_protocol`. Amazon MQ brokers sit behind a load balancer, so if it terminates the connection the peer is the LB and every caller collapses into one rate-limit bucket (default 10/60s becomes a global limit with no per-client isolation). `X-Forwarded-For` is not a safe fix (spoofable). The semaphore and body-size cap are the topology-independent backpressure. Worth confirming the Amazon MQ listener's source-IP behavior and, if the client IP is not preserved, either dropping per-IP limiting for an honest global token bucket or documenting the degradation.

**4. `dn_lookup_attribute` is accepted and shape-checked but never used in a directory check (low).** The broker uses it in an `equalityMatch` filter; the validator only confirms `dn_lookup_base` exists. A wrong attribute with a correct base still returns `ok`. The header documents this, but it is worth being explicit in user-facing docs so an `ok` is not over-trusted.

**5. Runtime enable without restart yields an uncaught 500 (low).** Workers start only at boot. If the env is flipped at runtime, `accept_content` calls `aws_auth_validate_rate_limiter:check/1` before the `try`, so the `gen_server:call` to an unregistered process exits and cowboy returns 500. A graceful 503/404 would be friendlier (operator-induced, so low priority).

**6. ARN-resolution lock is a `rabbitmq_aws` workaround (low).** `aws_auth_validate_arn_lock` serializes all ARN resolution because `rabbitmq_aws` holds region in global mutable state and resolves via a non-atomic `set_region`-then-call. The lock is a correct mitigation and should stay for this PR. A `rabbitmq_aws` fix that removes the global state is in progress; once the plugin adopts it, the lock and its `with_lock` wrapping can be removed (follow-up, not a blocker).

**7. Duplicate test module (low).** `aws_auth_validate_ldap_query_tests.erl` largely duplicates the query-parser tests already in `aws_auth_validate_tests.erl` (same corpus, parity test, and `literal_dns_*` cases). Consider removing one to avoid drift.

## Verified as correct (not issues)

- `binary_to_existing_atom` in the ssl-option translation is the right defense against atom-table exhaustion; unknown atoms are safely skipped
- No re-entrancy deadlock in the ARN lock (resolutions acquire it sequentially, never nested)
- The parser's top-level allowlist matches `rabbit_auth_backend_ldap_util:parse_query/1` exactly, while `literal_dns/1` still walks DN-bearing terms defensively
- `object_exists/2` mirrors the broker's `{exists,_}` probe
- The extension auto-mounts via the new `rabbitmq_management` dependency

src/aws_auth_validate_arn_lock.erl
sdewhitt: Making that edit right now, will update this PR later today

src/aws_auth_validate_ldap.erl
sdewhitt: Good catch, adding this right now

src/aws_auth_validate_mgmt.erl
sdewhitt: Sounds good, removing the rate limiter

src/aws_auth_validate_mgmt.erl
```diff
@@ -0,0 +1,298 @@
...
+                _Class:_Reason:_Stack ->
+                    audit(Method, SourceIP, connection_failed, T0),
+                    reply_error(
+                        400,
+                        connection_failed,
+                        <<"could not connect to LDAP server">>,
+                        Req,
+                        Context
+                    )
+            after
```
the-mikedavis, line 159: A catch-all like this resulting in "could not connect to LDAP server" might be too broad and not helpful to the end user. If we expect this to be basically unreachable, an internal server error code might be more appropriate.

I guess it's mainly a question of 'is this our error or the end-user's error?'

src/aws_auth_validate_ldap.erl
```diff
@@ -0,0 +1,623 @@
...
+        {versions, <<"versions">>, fun to_versions/1},
+        {server_name_indication, <<"server_name_indication">>, fun to_list/1}
+    ],
+    Translated = lists:foldl(
+        fun({SslKey, JsonKey, Fun}, Acc) ->
+            case maps:get(JsonKey, Map, undefined) of
+                undefined ->
+                    Acc;
+                Value ->
+                    case (catch Fun(Value)) of
```
the-mikedavis: Minor nitpick: `try`/`catch` is preferred now - the most recent Erlang versions deprecate `catch/1` as a keyword because of its error-swallowing tendencies.

the-mikedavis: I think this will catch and silently discard errors from `to_atom/1` below - specifically for `binary_to_existing_atom/2`? Would that discard a mis-typed option within build_ssl_opts?

I haven't looked into the test suite yet but it might be worth having a test where we explicitly give a mis-typed key and see a nice validation error caused by that.

src/aws_auth_validate_mgmt.erl
```diff
@@ -0,0 +1,298 @@
...
+                    audit(Method, SourceIP, capacity_exhausted, T0),
+                    reply_error(
+                        503,
+                        capacity_exhausted,
+                        <<"Validation service is not ready; broker restart required">>,
+                        Req0,
+                        Context
+                    );
+                true ->
+                    with_body(T0, SourceIP, Method, Req0, Context)
```
the-mikedavis: I think there's a TOCTOU here - small window where the check can say yes and then the semaphore _slightly_ later says no. I'm not super worried about this but it might be worth looking at refactoring this towards a pool-like/checkout implementation (I really like https://github.com/devinus/poolboy/blob/master/src/poolboy.erl) where you can atomically "check out" a worker/lease and then use it safely.

the-mikedavis (commented):
Continues to look good, just a few more minor things!

src/aws_auth_validate_mgmt.erl
sdewhitt: I've addressed the TOCTOU, but I decided not to do the pool approach since there shouldn't be any long-lived workers to pool due to each validation opening an ephemeral eldap connection that closes at the end of the validation request process. For this use case it seems like refactoring to the pool-like/checkout implementation could be excessive.

the-mikedavis (approved):
Nice, looks great to me now 👍

PR #52, opened 2026-06-17
Import `aws_lib` sources

lukebakken:
## Summary

Import the `aws_lib` sources into this repository as an additive, self-contained library, alongside the existing `rabbitmq_aws`-based code. Nothing in the plugin calls `aws_lib` yet; this PR lands the code, makes it build and test cleanly, and captures a backlog for follow-up work.

The motivation is that `rabbitmq_aws` keeps region and credentials in a single global gen_server and resolves ARNs via a non-atomic `set_region`-then-call, so concurrent callers can clobber each other's region. `aws_lib` removes that global state: state is an explicit `aws_state()` value threaded through every call, and `set_region/2` is pure. Adopting it in the plugin is intended as a follow-up, not part of this PR.

## What is in this PR

- Import the `aws_lib*` modules, the `aws_lib.hrl` header, their eunit tests and fixtures, and the BSD license files. The modules keep their original BSD headers; `NOTICE` records the provenance (BSD 3-Clause, originating from `rabbitmq_aws` in the RabbitMQ server tree, which came from httpc-aws). The existing Apache-2.0 `LICENSE` stays primary.
- Import the upstream README and agent guide as `README-aws_lib.md` and `AGENTS-aws_lib.md`.
- Add `gun` and `thoas` to `DEPS` so the generated `aws.app` declares them and dialyzer resolves the `gun:*` calls. Both are already present in the umbrella, so no version pin is needed.
- Capture 18 improvement notes for the `aws_lib*` modules under `docs/aws_lib/`, each with issue-ready front matter and an index, to be filed as issues after the plugin adopts `aws_lib`.
- Expand `config_schema_SUITE` to cover every `aws.schema` mapping. Removing the duplicate `aws_lib.schema` (which mapped `aws.prefer_imdsv2` into a second application environment and collided with the existing mapping) leaves that key resolving unambiguously.

## What is not in this PR

- No plugin module is switched from `rabbitmq_aws` to `aws_lib`
- `rabbitmq_aws` remains a dependency
- Reconciling how the `aws_lib` modules read `prefer_imdsv2` is deferred to the adoption work

## Testing

- `gmake` builds clean
- `gmake eunit`: 143 tests pass, including the imported `aws_lib_*` suites
- `gmake ct-config_schema`: 27 snippets pass, covering every `aws.schema` mapping
- `gmake xref`: clean (only benign dynamic dispatch)
- `gmake dialyze`: no new warnings from the imported code (pre-existing `aws_arn_config` and `aws_util` warnings are unchanged and out of scope)


sdewhitt (approved):
Looks good to me!

the-mikedavis (approved):
Overall this looks good and could probably be merged as-is.

It would be much more work but we could also consider rewriting from  scratch rather than starting with rabbitmq_aws source. A lot of the code hasn't aged well over time - use of strings rather than binaries, prop lists, etc.. But we could do that gradually too, starting with the core request modules.

PR #55, opened 2026-06-19
Feature: LDAP Validation Endpoint

sdewhitt:
*Issue #43*

*Forwarding PR #50, which merged from a fork to a feature branch in the public repo*

Allow customers to conveniently validate their LDAP broker configs via /api/aws/auth/validate/ldap with more descriptive (but safe) error messages instead of blindly guessing error causes, changing the config, and restarting the broker (~30 min per restart)

This endpoint is extensible for other forms of auth (HTTP/OAuth), however this PR focuses on solely implementing LDAP.

### Core subsystem (src/aws_auth_validate_*)
  - aws_auth_validate_mgmt -- Cowboy REST handler for PUT /api/aws/auth/validate/:method; pipeline: feature toggle → admin auth → body-size cap → JSON decode → concurrency semaphore → registry dispatch → audit log → response.
  - aws_auth_validate_registry -- method → backend dispatch with per-method enable flags and request-field filtering.
  - aws_auth_validate_backend -- behaviour (method_name/0, validate/1, allowed_fields/0).
  - aws_auth_validate_ldap -- LDAP backend: ephemeral eldap connect → optional StartTLS → simple bind → optional post-bind DN-lookup + authorization-query checks. All outcomes collapse to fixed response categories (no credential/LDAP detail leakage).
  - aws_auth_validate_ldap_query -- self-contained parser for the authz query DSL, mirroring rabbit_auth_backend_ldap_util:parse_query/1 exactly (R12 parity).
  - aws_auth_validate_semaphore -- supervised gen_server, started by aws_sup only when the feature is enabled.
  - aws_auth_validate_arn_lock -- supervised gen_server mutex that serializes ARN resolution. Meant to prevent region-clobbering with the global AWS credential state when resolving password/cacertfile ARNs via the validation endpoint.

### Config (priv/schema/aws.schema, aws_sup)
  - aws.auth_validation.* cuttlefish settings (toggle, body size, concurrency, timeout, user tag); workers conditionally supervised.
  - Defaults are applied in code, not declared as schema defaults, so they aren't emitted into every generated config (keeps config_schema_SUITE snippets exact).
  
### Tests
  - aws_auth_validate_tests (eunit): semaphore, ARN lock (value pass-through, serialization under concurrent callers, exception re-raise, survival of a crashing closure), registry, LDAP input validation, and query-parser + upstream-parity tests.
  - aws_auth_validate_mgmt_SUITE (CT): the HTTP pipeline (auth, body cap, semaphore, dispatch, response mapping) without a live LDAP server.
  - aws.snippets: config-schema coverage for the new settings.

### Response contract
  All failures return a fixed {"error": "<category>", "message": "<constant>"} -- never raw LDAP errors, hostnames, DNs, or the resolved password.
  Categories: input_invalid / connection_failed / tls_failed / query_invalid (400), auth_failed / config_conflict / authz_unverified (422), plus pipeline-level insufficient_user_tag (401), unknown_method / method_disabled (404), capacity_exhausted (503).

### Security notes
  - Endpoint disabled by default; admin-tag gated.
  - Ephemeral connections only -- no broker/auth-backend state mutation.
  - password_arn is resolved (via the plugin's existing ARN machinery) only after all pure input validation passes, so malformed requests never trigger a secret fetch. Resolution is serialized by aws_auth_validate_arn_lock so concurrent validations cannot clobber the shared client's region mid-resolution.
  - assume_role is deliberately not accepted (it would let a request swap the broker's global AWS credentials -- confused-deputy).


lukebakken (commented):
Review performed by AI (Claude) with @lukebakken guidance. Full review doc with verification status table available on request.

## Re-review of the LDAP Validation Endpoint

This is a follow-up to the review posted on PR #50. All seven findings from that review have been addressed:

1. OPTIONS preflight 401 on custom tag - fixed (`authorize_tag/3` explicit `user = undefined` clause + CT coverage)
2. Custom-tag branch untested - fixed (`feature_enabled_custom_tag` CT group)
3. `dn_lookup_attribute` shape-only - documented in module header
4. Runtime-enable yields HTTP 500 - fixed (atomic `try_acquire/0` + CT case)
5. Per-IP rate limiter behind LB - fixed (removed entirely; semaphore + body-size cap are topology-independent)
6. ARN lock workaround - improved (`global:trans/4`, no gen_server, R6 preserved)
7. Duplicate test module - fixed (removed)

## New findings (all low/informational - none blocking)

**1. SSRF filter DNS TOCTOU (low)** - `is_allowed_server/1` resolves the hostname and checks the IP, but `eldap:open` receives the original hostname string. A DNS rebinding attack could bypass the check if the attacker controls DNS between the two calls. The window is small and exploitation requires DNS control. Worth a comment; not blocking.

**2. CA-cert ARN resolution failure silently degrades TLS (high)** - When `cacertfile_arn` fails to resolve, `build_ssl_opts` gets `skip`, no `cacerts` entry is added, `apply_verify_default` finds no trust source, and the connection proceeds with effective `verify_none`. This contrasts with the password ARN path which returns a clear error. An operator whose CA cert is unreachable would not know their validation ran without certificate verification. Suggested: return `input_invalid` on the ARN-resolve failure, matching the password-ARN behavior.

**3. No HTTP-level 204 success path coverage (low, carried forward)** - `respond(ok, ...)` is never reached in the management CT suite. The 204 path is only exercised via direct `validate/1` calls in the LDAP integration suite.

**4. `LOCK_RETRIES = 600` effective timeout is opaque (informational)** - The `arn_lock_aborted` backstop is good, but there is no documentation of what 600 retries means in wall-clock time for an operator seeing it in logs.

## Verdict

The design is sound - pure-before-network ordering, R6 password-leak prevention, R12 query-parser parity, SSRF filter, upper-bound config validation, and the atomic semaphore liveness check are all well-executed. Finding 2 (silent CA-cert degradation) is the strongest candidate for a follow-up.

lukebakken (dismissed):
See item #2 ("CA-cert ARN resolution failure silently degrades TLS") in the comment the 🧞 and I posted.

src/aws_auth_validate_ldap.erl
```diff
@@ -0,0 +1,761 @@
...
+is_private_ip({127, _, _, _}) -> true;
+is_private_ip({10, _, _, _}) -> true;
+is_private_ip({172, B, _, _}) when B >= 16, B =< 31 -> true;
+is_private_ip({192, 168, _, _}) -> true;
+is_private_ip({169, 254, _, _}) -> true;
+is_private_ip({0, _, _, _}) -> true;
+is_private_ip(_) -> false.
+
+is_private_ip6({0, 0, 0, 0, 0, 0, 0, 1}) -> true;
+is_private_ip6({16#fe80, _, _, _, _, _, _, _}) -> true;
```
the-mikedavis: I think we also want `fc00::/7` (v6 private/ULA) here. This should block access to the IMDS-over-IPv6 address, `fd00:ec2::254`

the-mikedavis (commented):
One small thing, otherwise looks good to me

src/aws_auth_validate_ldap.erl
the-mikedavis, line 492: We could probably also block RFC6598 CGNAT (`100.64.0.0/10`) here

```
is_private_ip({100, B, _, _}) when B >= 64, B =< 127 -> true;
```

the-mikedavis (approved):

lukebakken (approved):
Clean merge. Post-connect peer check closing the SSRF TOCTOU is a nice touch.

priv/schema/aws.schema
```diff
@@ -95,6 +95,56 @@
...
+ end}.
+
+%% Note: none of the settings below declare a schema {default, ...}. A
+%% defaulted mapping is emitted into every generated config, which breaks
+%% config_schema_SUITE's exact-match snippets. Effective defaults are
+%% applied by the consuming code when the key is unset, so the key reaches
+%% the app env only when an operator sets it explicitly.
+
+%% @doc Maximum HTTP request body size in bytes (effective default 65536).
+{mapping, "aws.auth_validation.max_body_size", "aws.auth_validation_max_body_size",
```
sunfinite, line 130: What is the default max_body_size supported by the management plugin and why can't we rely on that?

priv/schema/aws.schema
```diff
@@ -95,6 +95,56 @@
...
+    [{datatype, integer}]}.
+
+%% @doc Maximum concurrent outbound auth-validation connections (effective
+%% default 5; applied by aws_sup).
+{mapping, "aws.auth_validation.max_concurrent", "aws.auth_validation_max_concurrent",
+    [{datatype, integer}]}.
+
+%% @doc Per-connection timeout in milliseconds for outbound calls
+%% (effective default 5000).
+{mapping, "aws.auth_validation.connection_timeout_ms", "aws.auth_validation_connection_timeout_ms",
```
sunfinite, line 140: Similar to my body size question, can't we rely on existing limits?

priv/schema/aws.schema
sdewhitt: The default from the management plugin is 20 MB, which is way overkill for the validation endpoint to rely on, so I scoped it down for this. The requirements table I have also says that the operator should be able to control the max body size as a P1

priv/schema/aws.schema
sdewhitt: I couldn't find an equivalent **outbound** connection timeout in the management plugin, so i implemented this myself

src/aws_auth_validate_ldap.erl
```diff
@@ -0,0 +1,829 @@
...
+%%--------------------------------------------------------------------
+
+%% ARN resolution mutates the shared rabbitmq_aws region/credential
+%% singleton (aws_sms/aws_acm_pca call rabbitmq_aws:set_region/1), so
+%% serialize it across concurrent validation requests to prevent region
+%% clobbering between a set_region and its HTTP call. See
+%% aws_auth_validate_arn_lock.
+resolve_arn(Arn) when is_binary(Arn) ->
+    aws_auth_validate_arn_lock:with_lock(fun() ->
+        aws_arn_util:resolve_arn(binary_to_list(Arn))
```
sunfinite, line 1052: I see we chose not support the assume_role setting, so would this call not use the wrong creds if `assume_role_arn` is configured. For reference, in the boot sequence of the plugin, we assume the configured role, resolve all ARNs and then reset creds.

src/aws_sup.erl
```diff
@@ -16,10 +16,57 @@ start_link() ->
...
+        id => aws_auth_validate_semaphore,
+        start => {aws_auth_validate_semaphore, start_link, [Config]},
+        restart => permanent,
+        shutdown => 5_000,
+        type => worker,
+        modules => [aws_auth_validate_semaphore]
+    }.
+
+semaphore_config() ->
+    #{max => get_int_env(auth_validation_max_concurrent, 5, 100)}.
```
sunfinite, line 60: Should this be 1 by default until we iron out the multi creds issue?

src/aws_auth_validate_ldap_query.erl
```diff
@@ -0,0 +1,188 @@
...
+            end
+    end.
+
+scan(Str) ->
+    case erl_scan:string(Str) of
+        {ok, Tokens, _EndLine} -> {ok, Tokens};
+        {error, _Info, _Loc} -> {error, scan_failed}
+    end.
+
+%% Top-level allowlist -- mirrors rabbit_auth_backend_ldap_util:parse_query/1.
```
sunfinite, line 82: Why do we need to mirror and not call the plugin function directly?

src/aws_sup.erl
sdewhitt: I modified the semaphore's behavior to not be gating concurrent access to the aws state, but rather to be a guard on how many validation requests are processing at a time

src/aws_auth_validate_ldap_query.erl
```diff
@@ -0,0 +1,188 @@
...
+%% This intentionally mirrors the grammar accepted by
+%% rabbit_auth_backend_ldap_util:parse_query/1 (erl_scan + erl_parse:parse_term
+%% over the same accepted-term allowlist) so that a query the broker would
+%% accept is accepted here and vice-versa (design requirement R12: the
+%% validation endpoint must not diverge from rabbit_auth_backend_ldap). A
+%% parity test in aws_auth_validate_tests guards against drift.
+%%
+%% Two deliberate differences from the upstream helper:
+%%   1. It returns {ok, Query} | {error, Reason} instead of throwing via
+%%      cuttlefish:invalid/2 -- this runs in the request path, not at
```
sunfinite, line 18: Can't we catch the cuttlefish:invalid throw?

src/aws_auth_validate_ldap_query.erl
```diff
@@ -0,0 +1,188 @@
...
+%%      config-load time.
+%%   2. It adds literal_dns/1, which walks a parsed query and returns the
```
sunfinite, line 20: Can we remove this limitation if we accept a username in the API request? I'm not sure how useful evaluating queries with no args in reality.

src/aws_auth_validate_ldap.erl
sdewhitt: I missed that the creds would be reset at the end of the boot sequence. I had assumed that the broker creds themselves would be satisfactory for ARNs in the validation, so Beichen and I decided based off that that it would be safe to sidestep the global state credential issue by not using assume_role. I currently have a PR #68 in progress to integrate per-request credentials, and use assume_role securely, but that is not present in this PR

src/aws_auth_validate_ldap_query.erl
sdewhitt: cuttlefish:invalid/2 doesnt throw, it instead raises erlang:error({invalid, _}), which is not supposed to be catched. It's cleaner just to return {ok, Query} | {error, Reason}

src/aws_auth_validate_ldap_query.erl
sdewhitt: The main reason is managing dependencies. rabbitmq_auth_backend_ldap is currently just a TEST_DEPS, not a runtime DEPS. While we want to use the functionality of parse_query, if we don't mirror it in this plugin, then we need to add rabbitmq_auth_backend_ldap as a runtime DEPS.

I deemed it wasn't worth adding on the entire auth backend just to only use ~40 lines of code

src/aws_auth_validate_ldap.erl
sdewhitt: I did some more digging, and this approach is still somewhat valid. #68 is still relevant, however after the creds are reset in the boot sequence, when a user sends a request to the validation endpoint, the instance/task IAM role that the broker's EC2 instance uses will refresh and provide creds for ARN resolution. This prevents the endpoint from being totally defunct, but there IS the caveat that the ARNs can only be resolved if scoped within the instance role's perms

sdewhitt: Flagged as stretch goal in #75 to prevent internship scope creep

sunfinite: To clarify, there is already an assume_role_arn in the configuration which this plugin should always use when present. We should not use instance role creds if the plugin is configured to use another role. The stretch goal would be remove this fallback to instance role creds :)

lukebakken (approved):
Approving. Fresh-eyes re-review of the full diff plus all prior review threads.

The architecture is sound and notably careful: pure-before-network ordering, R6 password-leak prevention (try/catch discards class/reason/stack so the bind password cannot reach a crash report), R12 query-parser parity, the post-connect peer re-check that closes the SSRF DNS-rebinding TOCTOU, and the monitor-based semaphore. The earlier CA-cert TLS-degrade finding is fixed (resolve/decode failures return input_invalid, never a silent verify_none), LDAP injection via username/user_dn is not exploitable (ASN.1-encoded equalityMatch, and the filled user_dn is the directory-derived object_name, not caller input), and no secret or LDAP detail leaks into responses.

Approving with the explicit understanding that #68 (aws_lib integration) merges right after this. #68 supersedes a few items here, so they are not blockers for this PR: the untrusted query string reaching erl_scan:string/1 (atom-table exhaustion DoS) is replaced by the non-interning lexer, the status_for_category catch-all is added, and the SSRF blocklist gains NAT64 and 240.0.0.0/4. These are pre-release and internal-only in the window between the two merges.

Two things for whoever merges #68 on top:

- assume_role: the validation endpoint resolves password/cacert ARNs under the node instance-role creds, not the configured assume_role_arn, and falls back silently. This is a least-privilege and broker-parity gap that #68 does not fix either. Tracked as the stretch goal in #75; fine to defer, but it should not be left undocumented.
- #55 and #68 are divergent branches editing the same files, and #68 does not contain this PR's username live-membership feature (no username in allowed_fields, no resolved_user_dn/fill path). The #68 merge must consciously reconcile so the username feature added here is preserved rather than silently dropped.

PR #66, opened 2026-06-26
Feature: HTTP Validation Endpoint

sdewhitt:
*Issue #43*

Allow customers to conveniently validate their HTTP broker configs via /api/aws/auth/validate/http with more descriptive (but safe) error messages instead of blindly guessing error causes, changing the config, and restarting the broker.

This endpoint extends off of the validation endpoint introduced in PR #55, with this PR focusing on HTTP. The shared framework (Cowboy handler, registry, concurrency semaphore, backend behaviour, config schema, audit logging) is reused unchanged; this PR adds the HTTP backend, its SSRF address policy, and mutual-TLS support.

**Scope: reachability-only.** Unlike an LDAP bind, an HTTP auth server answers `allow`/`deny` for a specific {user, vhost, resource} tuple -- a `deny` is a correctly-working server, not a misconfiguration. So this validates **reachability, (m)TLS, and response shape**, not authorization outcome: it confirms each configured `*_path` is a well-formed http(s) URL, is reachable over (m)TLS, and answers with a body matching `rabbit_auth_backend_http`'s allow/deny grammar. It checks the **shape** of the response, not its verdict -- a `deny` for the synthetic probe principal is a success, since a well-formed deny still proves the endpoint speaks the auth protocol. A credentialed-probe mode (asserting a *specific* principal is allowed) is intentionally deferred.

### Core subsystem (src/aws_auth_validate_http)
  - aws_auth_validate_http -- the HTTP backend (aws_auth_validate_backend behaviour: method_name/0, validate/1, allowed_fields/0). Pipeline: pure input validation (URL/method/ssl_options shape) → SSRF address guard → ARN resolution (cacertfile/certfile/keyfile) → ephemeral httpc probe over (m)TLS → response-contract check → fixed-category result.
  - **Response-contract check:** a usable status (200/201) is necessary but not sufficient -- the body must also match `rabbit_auth_backend_http`'s grammar (normalize via `lower(strip(body))`, then `user_path` accepts `deny` or `allow`-with-optional-tags; the authz paths accept a bare `allow`/`deny`). This closes the false-pass where a path pointing at any 200-returning service (health check, HTML page, proxy) would otherwise read as a working auth backend. A 2xx whose body is neither allow nor deny is `auth_failed`.
  - URL parsing rejects non-http(s) schemes, out-of-range ports, and userinfo (`user:pass@host`).
  - Accepted request fields mirror the customer's `auth_http.*` config so a config can be pasted as-is: `user_path` (required), `vhost_path`/`resource_path`/`topic_path` (optional), `http_method` (get/post), and `ssl_options` (`verify`, `depth`, `versions`, `sni`, `cacertfile_arn`, `certfile_arn`, `keyfile_arn`).
  - Mutual TLS: `certfile_arn` (S3-hosted client cert PEM) + `keyfile_arn` (Secrets Manager client key PEM) are resolved and decoded into in-memory ssl `{cert,_}`/`{key,_}` options. `verify_peer` with no obtainable trust anchor fails loudly (`tls_failed`) rather than silently downgrading to `verify_none`.

### SSRF address policy (the central security concern for HTTP)
  Unlike LDAP, the validator issues requests to customer-supplied URLs from the broker's network position, so SSRF is front-and-center. Two-phase guard:
  - **Pure phase:** scheme allowlist (http/https) + literal-IP classification against a broker-infra denylist, before any DNS or network.
  - **Network phase:** resolve the host to all A/AAAA records, deny if any is infra, then connect to the **pinned IP literal** (original hostname kept only in the Host header + SNI) so httpc cannot re-resolve to a different address -- the DNS-rebinding/TOCTOU defense.
  - Denylist is **broker-infra only** (IMDS 169.254.169.254 + fd00:ec2::254, loopback, link-local, unspecified), with v6-embedded-v4 unwrapping (mapped/compatible/NAT64/6to4) so a v6 encoding cannot smuggle a denied v4. RFC1918 / unique-local are **deliberately allowed** -- a customer's HTTP auth server normally lives in their VPC. Redirects are not followed.

### Config (priv/schema/aws.schema, aws_sup, registry)
  - No new config schema or supervised workers: the HTTP method reuses the existing `aws.auth_validation.*` settings and the concurrency semaphore.
  - The `http` method is **opt-in** in the registry (defaults disabled even when `ldap` is enabled), so enabling validation for LDAP does not silently bring the HTTP probe online.

### Tests
  - aws_auth_validate_http_tests (eunit, 73 tests): behaviour callbacks; pure input validation (every fixed reason -- paths, URL, scheme, http_method, ssl_options shape/keys/values, cert-without-key); SSRF classify_ip (IMDS/loopback/link-local denied, RFC1918/public allowed, v6-embedded-v4 unwrap denied); ARN-first ordering (resolve not called when pure validation fails); the response-contract grammar (allow/deny/allow-with-tags accepted, HTML/JSON/empty/near-miss bodies rejected, whitespace+case normalization, authz exact-match); and R6 no-leak (resolved secret never appears in a rendered error, including the probe-raise path).
  - aws_auth_validate_http_SUITE (Common Test, 12 cases): an in-process inets stub HTTP(S) server exercises the real httpc probe end to end -- the probe outcomes (200/201 with an allow/deny body → ok, a 200 with a non-auth body → auth_failed, 4xx/5xx → auth_failed, closed port → connection_failed), self-signed cert under `verify_peer` → tls_failed and under `verify_none` → ok, the `verify_peer` + customer-CA (`cacertfile_arn`) success path as a regression guard that the CA bundle is passed to ssl as raw DER (decoded records are silently ignored → unknown_ca), and **end-to-end mutual TLS**: against a stub that requires a client cert (`verify_peer` + `fail_if_no_peer_cert`), presenting `certfile_arn`/`keyfile_arn` → ok, and withholding them → tls_failed (proving the client cert is actually wired through to ssl, not just accepted as input). The suite relaxes only loopback for its local stub via `auth_validation_allow_private_networks`; production address policy is unchanged.
  - Internal helpers are exposed under `-ifdef(TEST)` for direct unit coverage, mirroring the LDAP backend.

### Response contract
  Reuses the existing fixed category set -- **no new category** is introduced. All failures return `{"error": "<category>", "message": "<constant>"}`, never a raw httpc error, target URL, response body, or resolved secret.
  HTTP-relevant categories: input_invalid / connection_failed / tls_failed (400), auth_failed (422, borrowed for "reachable but not auth-server-shaped"), plus the inherited pipeline-level categories. 204 on success.

### Security notes
  - Method disabled by default and opt-in even when other methods are enabled; admin-tag gated.
  - Ephemeral connections only, on a dedicated per-validation httpc profile (no pooled/reused TLS sessions, so one validation's client identity cannot leak into the next). No broker/auth-backend state mutation.
  - ARNs (cacertfile/certfile/keyfile) are resolved only after all pure input validation passes, so malformed requests never trigger a secret fetch.
  - `assume_role` is deliberately not accepted (confused-deputy on the broker's global AWS credentials), consistent with #55.

### Excluded from this PR
  - **Credentialed / authorization-outcome probing** (asserting a principal is allowed) -- needs a real test principal and a definition of "expected outcome"; deferred.
  - **OAuth** -- a separate method/PR.
  - **CGNAT (100.64.0.0/10) blocking** -- the LDAP backend blocks all private/shared ranges by default (a directory is not public-facing); the HTTP backend deliberately blocks only broker-owned infra and allows customer-VPC ranges, so CGNAT (customer/provider space reachable by legitimate auth servers) is intentionally allowed. This asymmetry follows from the two backends' opposite denylist policies, not an oversight.


lukebakken (commented):
Thorough initial pass on the HTTP backend, the two supporting changes, and the shared framework it plugs into. Verification: read all five changed files plus `aws_auth_validate_backend`/`_registry`/`_mgmt`/`_semaphore` and the LDAP backend for parity; ran eunit (73/73), the CT suite (12/12, exit 0), xref (no new breakage), and dialyzer.

This is high-quality, security-conscious work. The two-phase SSRF guard, ARN-after-pure-validation ordering, the R6 no-leak discipline, the ephemeral-profile isolation, and the fail-loud no-trust-anchor policy are all sound, and the response-contract check that closes the "any 200 reads as an auth server" false pass is a genuinely good addition. Findings below, by severity.

## High

### H1: dead fail-closed branch, and the one dialyzer regression this PR introduces

`ssrf_policy_ready/0` returns `?ENFORCE_ADDRESS_POLICY`, which is now hardcoded `true`, so the `false -> {error, method_disabled}` arm of `validate/1` is unreachable. Dialyzer flags exactly this:

```
aws_auth_validate_http.erl:222:9: The pattern 'false' can never match the type 'true'
```

I confirmed this is the *only* new dialyzer warning. Every other warning in the run is pre-existing baseline: `aws_auth_validate_mgmt.erl` is untouched by this PR yet carries several, and the registry `get_env` warning traces back to the PR #55 commit e94e192. CI does not gate dialyzer today (build-test runs `make` + `make tests` only), so this will not fail the build, but it is a real dead branch.

Separately, `{error, method_disabled}` is a 2-tuple that is not part of the behaviour's `result() :: ok | {error, error_category(), binary()}` type. It only works because `aws_auth_validate_mgmt` happens to also match the 2-tuple form. Now that the address policy is live, I would drop `ssrf_policy_ready/0` and the outer `case` in `validate/1` entirely (the opt-in registry gate is the actual runtime switch), which also clears the dialyzer error.

### H2: the profile-pool collision analysis is inverted, and a collision produces a spurious failure

The comment above `?PROFILE_POOL_SIZE` argues the 128-slot pool sits "well above `auth_validation_max_concurrent`'s cap (100), so concurrently-running validations almost never collide." The cap of 100 is real (`aws_sup.erl:61`), but the conclusion is backwards. With up to 100 concurrent validations each choosing `erlang:phash2(make_ref(), 128)`, the probability that some pair collides is effectively 1 (birthday problem), not rare.

And a collision is not benign. `start_probe_profile/1` handles `{already_started, _}` by calling `inets:stop(httpc, Profile)` on a slot a *concurrent* validation may be actively probing on, tearing down its in-flight request. The victim then returns `connection_failed` for a customer auth server that is actually reachable: a false negative handed to a legitimate operator, and the exact "your server is unreachable" misreport this endpoint exists to avoid.

Since concurrency is already bounded by the semaphore, the clean fix is to derive the profile identity from the acquired semaphore slot (or otherwise pass a per-request unique-but-bounded token through), so two in-flight validations can never share a profile by construction. That keeps the atom table bounded and removes the collision entirely.

## Medium

### M1: the DNS-rebinding defense has no end-to-end coverage

All 12 CT cases target the literal `127.0.0.1`, which takes the literal-IP branch of `resolve_and_pin/1`. The hostname branch, `resolve_hostname_and_pin/2` (resolve to all A/AAAA, deny if any is infra, pin to the resolved IP while keeping the original name in the Host header and SNI), is the central SSRF control for hostname targets and is never exercised by the suite. `classify_ip/1` is unit-tested on address lists, but the pin-and-rewrite path is not. Suggest one CT case with a hostname target (for example `localhost` under `auth_validation_allow_private_networks`) that asserts the stub receives the rewritten Host header.

### M2: `ensure_trust_anchor/1` and `trust_source/1` are identical

Both functions have byte-for-byte the same body (keymember `cacerts` then `{ok, Opts}`, else fall back to `os_cacerts/0`). Collapse to one.

## Low

### L1: unused `base_body/0`

The compiler warns that `base_body/0` is unused in `aws_auth_validate_http_tests.erl`; only `base_body/1` is called. Remove the arity-0 clause.

### L2: `parse_url/1` rejects a query and userinfo but not a fragment

A `*_path` carrying a `#fragment` survives parsing. `build_request(get, ...)` then appends `?<query>` after the fragment, yielding `.../path#frag?username=...`, where the query is absorbed into the fragment and never sent, so the probe silently misfires and reports `auth_failed`. Auth paths do not carry fragments in practice, but rejecting a fragment in the same clause that already rejects a query is a cheap robustness win.

### L3: `max_sessions => 0` semantics are unverified

I could not confirm that `0` is a meaningful value for httpc's `max_sessions` option. The real isolation comes from the fresh profile plus its teardown in the `after` clause, so `set_probe_profile_opts/1` is belt-and-suspenders, but a quick confirmation that `0` is honored (not silently ignored or clamped) would be worth it.

## Summary

Solid design, no data-loss or secret-leak concerns. H1 and H2 are the two I would want addressed before this leaves draft: H1 is a trivial removal, H2 is the one with real user-facing impact under concurrency. Everything else is polish.

lukebakken (approved):
Re-reviewed against the latest branch. Ran dialyzer, eunit (76/76), and the CT suite (14/14), and traced the httpc/inets and supervisor internals in OTP source to settle the two open questions from the first review.

All six first-review findings are resolved:

- **H1** (dead fail-closed branch + dialyzer regression): the guard and `ssrf_policy_ready/0` are gone; dialyzer now reports zero warnings for the module
- **H2** (profile-pool collision + slot-stealing): replaced by `claim_probe_profile/2`, a bounded scan that only claims a slot by starting it and advances rather than stopping a peer's slot. The invariant holds in source: `max_concurrent` is clamped to 100 in aws_sup.erl, below `?PROFILE_POOL_SIZE` (128), so one pass always finds a free slot
- **M1** (no DNS-rebinding coverage): two new CT cases exercise `resolve_hostname_and_pin/2` end to end, including the pin+Host-rewrite and the deny-on-infra path
- **M2** (duplicate trust-anchor helpers): `trust_source/1` removed
- **L1** (unused `base_body/0`): removed
- **L2** (`parse_url/1` accepted a fragment): now rejected, with a test

Two points I had left open are also settled against OTP source:

- **`max_sessions => 0`**: confirmed honored. `httpc:validate_max_sessions/1` accepts `>= 0`, and `httpc_manager` forces `Connection: close` with no persistent connection when it is 0, so session pooling is genuinely disabled
- **httpc vs gun**: the httpc choice is correct, not an inconsistency. `rabbit_auth_backend_http` (the backend this validates) issues its own requests via `httpc:request/4`, so probing with the same client is what makes a green result predictive of production. Mirrors the LDAP validator using eldap

One robustness gap surfaced while tracing the H2 loop against the supervisor internals: `inets:stop/2` tears a profile down as `terminate_child` then `delete_child`, and a permanent child keeps its spec (`pid=undefined`) between them, so a claim landing in that window gets `{error, already_present}` rather than `{error, {already_started, _}}` and fell through to the fail-closed catch-all, returning `connection_failed` on the ordinary concurrent stop/claim path. It fails safe (no false success or leak), so it was never a correctness or security hole. Now handled: `already_present` advances like `already_started`, covered by a test.

No outstanding concerns. No data-loss or secret-leak issues. Looks good to me.

src/aws_auth_validate_http.erl
```diff
@@ -0,0 +1,1191 @@
...
+%% so concurrent validations no longer share mutable region/credential state and
+%% no ARN lock is needed. R6 is preserved -- resolution runs in the caller
+%% process and the resolved secret is neither logged nor returned (only adapted
+%% to the caller's {ok, Binary} contract). R3 is preserved -- this still runs
+%% only after the pure validation pipeline. The 3-tuple {ok, Data, State1} from
+%% aws_arn_util:resolve_arn/2 is adapted back to the {ok, Binary} contract the
+%% callers expect; the threaded state is request-scoped and discarded here.
+-spec resolve_arn(binary()) -> {ok, binary()} | {error, term()}.
+resolve_arn(Arn) when is_binary(Arn) ->
+    State = aws_lib:new(),
```
the-mikedavis: Will this get the same treatment in the future as #109?

the-mikedavis (approved):
Looks good overall. Just one non-blocking question.

src/aws_auth_validate_http.erl
sdewhitt: yes! I would like to merge #109 first and then patch this in

sdewhitt: Just committed a patch for it onto this PR

the-mikedavis (approved):
🚀

PR #68, opened 2026-06-26
Integrate aws_lib, moving away from rabbitmq_aws

sdewhitt:
*Issue #53*

aws_lib was previously imported into the plugin in #52 but left dormant. This PR activates it as the plugin's actual AWS client, replacing `rabbitmq_aws` and sidestepping the shared global AWS state it relied on (region-clobbering between concurrent resolutions, `assume_role` risk, and the serialization lock those forced).

The core change is mechanical but pervasive: ARN/credential resolution moves from a **mutated global singleton** to **per-call state threading** (`aws_lib:new()` + `resolve_arn/2`'s `{ok, Data, State}` contract). Because each resolution carries its own state, concurrent validations can no longer clobber one another's region, which removes the need for the lock that #55 introduced to work around exactly that.

> **Stacking note (please read before reviewing the diff):** this branches off `feature/ldap-validation-endpoint` (PR #55), which is not yet merged to `main`. The diff against `main` therefore also contains the full LDAP endpoint from #55. **Only the aws_lib wiring below is new in this PR**; the `aws_auth_validate_*` endpoint itself is reviewed in #55. The history deliberately carries no HTTP-validation-endpoint commits (#66).

### AWS client migration (src/aws_arn_util, aws_s3, aws_sms, aws_iam, aws_acm_pca, aws_arn_config*)
  - Route all ARN/credential resolution through aws_lib: each module now threads aws_lib state per call instead of reading/mutating the `rabbitmq_aws` global region/credential singleton.
  - `aws_arn_util:resolve_arn/2` returns `{ok, Data, State}` so callers carry state forward rather than relying on process-global config.
  - `aws_arn_config*` (including the OAuth2 cacertfile path) updated to the threaded-state contract.

### Auth-validation endpoint wiring (src/aws_auth_validate_ldap)
  - The LDAP backend resolves `password_arn` / `ssl_options.cacertfile_arn` via the aws_lib path.
  - **Removed `aws_auth_validate_arn_lock`** -- the gen_server mutex that serialized ARN resolution to prevent region-clobbering is unnecessary once state is threaded per call. This deletes a supervised process and its `aws_sup` wiring.

### Validate endpoint honors the configured assume role (new)
  - The validate endpoint now resolves its ARNs (`password_arn`, `ssl_options.cacertfile_arn`) under the operator-configured `aws.arns.assume_role_arn` when one is set, instead of falling back to the broker's bare ambient (EC2 instance) role. This is the **same role the plugin already assumes at boot** to resolve every configured ARN (`aws_arn_config:maybe_assume_role/1`), so the endpoint now resolves ARNs the same way the rest of the plugin does -- closing a parity gap where a config readable at boot would fail to resolve through the endpoint.
  - `resolve_request_state/1` builds one per-request aws_lib state, assuming the configured role into it (`aws_iam:assume_role/2`) or using a default state when none is configured; `configured_assume_role_arn/0` reads the role from the same `aws, arn_config` env the boot path uses. The state is threaded through `resolve_password/2`, `resolve_cacert/2`, and `resolve_arn/2`.
  - This is **operator configuration, not caller input** -- the role is read from `rabbit.conf`, never from the request body -- so it raises no confused-deputy concern. The assumed credentials live only in the request-local aws_lib value (no global singleton), so concurrent validations cannot affect one another or the broker. A failed assume surfaces as `input_invalid` in the network phase, after all pure validation, preserving ARN-after-validation ordering. With no role configured, behavior is byte-for-byte the prior ambient-role path.

### Config & supervision (priv/schema/aws.schema, aws_sup)
  - `aws_sup` and the cuttlefish schema/snippets updated for the aws_lib wiring (one fewer supervised worker now that the ARN lock is gone).
  - Minor `aws_lib_config` refinements with accompanying tests.

### Cleanup
  - Removed the now-unused `aws_util`.
  - Removed the unwired literal-DN probe budget (`limit_literal_dns`/`?MAX_LITERAL_DNS`): it was defined but never added to the parse pipeline, so it was dead code that failed the warnings-as-errors build. Re-adding it as a real backpressure guard is tracked as follow-up.

### Tests
  - `aws_lib_config_tests` extended for the config refinements.
  - `ldap_configured_assume_role_test_` covers the new fallback by mecking `aws_iam:assume_role/2` + `aws_arn_util:resolve_arn/2` and asserting the three branches: not configured (ambient credential-free state reaches resolution; `assume_role` never called), configured+success (the assumed credentials reach resolution), and configured+failure (`input_invalid` before any ARN resolve). Mirrors the boot-path coverage in `aws_arn_config_tests`.
  - The existing eunit + CT suites (semaphore, registry, LDAP input validation, query parser/parity, HTTP pipeline) pass unchanged against the threaded-state path -- the migration is behavior-preserving at the response-contract level.
  - Verified: `gmake` build clean; `erlfmt -c` clean; CT 18/18 against slapd; eunit green (one pre-existing, unrelated `aws_arn_config_tests:replace_in_env_test` failure -- an OS-trust-store artifact, not this change).

### Security notes
  - Eliminates the shared-global-state class of bug by construction: with per-call state there is no mutable region/credential singleton for one request to clobber for another, so the serialization lock is no longer load-bearing.
  - **Configured** `assume_role_arn` (operator-set, from `rabbit.conf`) is now honored by the validate endpoint, matching boot behavior. **Request-supplied** `assume_role` (a role in the request body) remains deliberately unaccepted -- that would be a confused-deputy on the broker's AWS credentials, consistent with #55 and #66.
  - No change to the endpoint's response contract, ARN-first ordering, or no-leak guarantees -- ARNs are still resolved only after pure input validation passes.

### Excluded from this PR
  - **HTTP / OAuth validation backends** -- separate methods/PRs (#66 and follow-on).
  - **Request-supplied `assume_role` in the validate API call** -- a possible stretch feature (tracked in #75), gated on an operator allowlist and AppSec review; not in scope here.

closes #53


lukebakken (commented):
@sdewhitt — fresh-eyes security pass. Five items below; I (@lukebakken) will address all five in follow-up commits.

**1. Atom-exhaustion DoS — `aws_auth_validate_ldap_query.erl:68` (high)**
`parse/1` runs `erl_scan:string/1` on the raw query before `accept/1`, interning a persistent atom per unique identifier. Authenticated, but a few requests at the 64KB body cap can approach the ~1M atom limit. Bound identifier/token count before scanning.

**2. OAuth2 handler drops threaded state — `aws_arn_config_oauth2.erl:32` (medium)**
`run/4` discards `_FinalState` and returns bare `ok`, so `run_arn_handlers/2` continues with the pre-oauth2 state. Impact is a redundant credential refresh and a broken threading contract (not expired creds). Return `{ok, FinalState}` and thread it.

**3. Schema bounds on security knobs — `priv/schema/aws.schema:132-148` (medium)**
`max_body_size`, `max_concurrent`, `connection_timeout_ms` are bare `{datatype, integer}`; out-of-range values silently fall back to defaults at runtime. Add `{min,_}/{max,_}` (cf. `rabbit.schema`) so cuttlefish rejects them at config time.

**4. Missing `enabled = false` schema snippet — `test/config_schema_SUITE_data/aws.snippets` (low)**
Only `= true` is covered; add the `false` case (as done for `aws_prefer_imdsv2`).

**5. No catch-all in `status_for_category/1` — `aws_auth_validate_mgmt.erl:191` (low)**
Theoretical today (only the LDAP backend dispatches, all categories map; unmapped ones hit the outer 500 catch). Add an explicit catch-all for defense in depth.

the-mikedavis (approved):
I don't know if we need to block this PR on aws_lib improvements, but I think there are some behavior changes in failure modes between httpc and gun (HTTP client libraries). In `aws_lib` it looks like there's `error({gun_connection_failed, Reason})` / `error({gun_open_failed, Reason})` and I think that will escape the retry logic. And similarly for `gun:await_body` in `aws_lib_config` I believe a timeout (`{error, timeout}`) would be a badmatch, where with httpc it would be returned as `{error, timeout}`.

Otherwise I think this looks good 👍

lukebakken (approved):
Approving. Thorough re-review of the full diff plus the gun-vs-httpc failure-mode points @the-mikedavis raised, both of which are now addressed:

- 8e1f1f4 fixes the more serious one: a connection that could not be established (DNS failure, connection refused, TLS handshake failure, await_up timeout) was raised out of create_gun_connection/3 and escaped the retry loop in api_request_with_retries/8 entirely, where the old httpc client returned {error, Reason} and got retried. create_gun_connection/3 now returns {ok, ConnPid} | {error, Reason}, routed through format_response/1 into the {error, _, _} shape the retry loop matches; open_connection/3 honors its existing {error, term()} spec. This also surfaced a latent bug fixed in the same commit: the retry-loop warning logged the message with ~ts, which crashed once the message became a {gun_open_failed, _} tuple, so it is now ~tp. The two new tests were confirmed to fail against the pre-fix raising code with the exact escaping exceptions.
- 97c2cb8 fixes the second: a {error, timeout} from gun:await_body/3 badmatched and the surrounding catch turned it into {error, {badmatch, {error, timeout}}}. It is now matched explicitly and returned as a clean {error, timeout}, matching the prior httpc behavior.

Also pushed earlier on this branch:

- 2addf98 makes the slapd test harness probe /usr/lib64/openldap, so aws_auth_validate_ldap_SUITE runs on Amazon Linux 2023 (and other 64-bit RH-family hosts) instead of skipping with "slapd unreachable".

Verification: full suite (eunit + CT, including aws_auth_validate_ldap_SUITE against a real slapd) runs green locally at 298 tests; gmake xref and gmake dialyze are clean; the atom-exhaustion fix was verified non-interning against compiled code (zero atom-table growth across 100/2000/8000 distinct identifiers). CI is green across all four matrix legs.

PR #105, opened 2026-06-30
Treat all 2xx status codes as success in `format_response`

lukebakken:
Fixes #83.

`format_response/1` only matched status codes 200, 206, and >= 400. A 201, 202, or 204 success (returned by S3 PUT, DynamoDB writes, and other AWS APIs), as well as any 3xx, fell through every clause and crashed the caller with a `function_clause` error.

This matches any 2xx as success with a range guard and treats every other status as an error. gun is not configured to follow redirects, so a 3xx is a request we could not complete and is correctly reported as an error.

Tests cover 201, 204, and a 3xx redirect in `format_response_test_`; they were confirmed to fail with `function_clause` against the previous code. `gmake eunit` and `gmake xref` are clean.

### DIFF
--- a/src/aws_lib.erl
+++ b/src/aws_lib.erl
@@ -518,11 +518,14 @@ do_refresh_credentials(State0) ->
 %% structure. The response body will attempt to be decoded by invoking the
 %% maybe_decode_body/2 method.
 %% @end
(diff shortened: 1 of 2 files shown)


sdewhitt (approved):
Looks good to me!

PR #106, opened 2026-06-30
Handle `gun:await_body` errors during credential loading

### LINKED ISSUE #90: `aws_lib`: gun:await_body timeout crashes aws_lib_config with badmatch
## Problem

In `aws_lib_config.erl`, calls to `gun:await_body/3` use assertive pattern matching:

```erlang
{ok, Body} = gun:await_body(ConnPid, StreamRef, ?DEFAULT_HTTP_TIMEOUT),
```

`gun:await_body/3` can return `{error, timeout}` or `{error, Reason}` if the body delivery times out or the connection drops after headers have been received. These calls are NOT inside a try/catch block, so a timeout on body delivery crashes with `badmatch`.

This affects:
- `perform_http_get_with_conn/3`
- `perform_http_get_instance_metadata/2`
- `load_imdsv2_token/0`

## Impact

If the EC2 Instance Metadata Service responds with headers (200 status) but the body delivery is delayed beyond 2250ms, credential loading crashes instead of returning `{error, timeout}`.

## Fix

Handle the error case from `gun:await_body`:

```erlang
case gun:await_body(ConnPid, StreamRef, ?DEFAULT_HTTP_TIMEOUT) of
    {ok, Body} -> ...;
    {error, Reason} -> {error, Reason}
end
```

lukebakken:
Fixes #90.

Three call sites in `aws_lib_config` matched the body read with `{ok, Body} = gun:await_body(...)`. `gun:await_body/3` can return `{error, timeout}` (or another `{error, Reason}`) when body delivery stalls after the response headers arrive, so a slow EC2 Instance Metadata Service response crashed credential loading with a badmatch instead of failing gracefully.

Each site now matches `await_body/3` explicitly and propagates the error per the function's contract:

- `perform_http_get_with_conn/3` returns `{error, Reason}`
- `perform_http_get_instance_metadata/2` returns `{error, Reason}`
- `load_imdsv2_token/0` logs and returns `undefined`, so it still falls back to IMDSv1 (this function is spec'd to return a token or `undefined`, not an error tuple)

Tests cover the body-read-timeout path for the metadata-credentials and IMDSv2-token requests; both were confirmed to fail with `{badmatch, {error, timeout}}` against the previous code. `gmake eunit` and `gmake xref` are clean.

### DIFF
--- a/src/aws_lib_config.erl
+++ b/src/aws_lib_config.erl
@@ -765,8 +765,13 @@ perform_http_get_with_conn(ConnPid, Path, Config) ->
         {response, fin, Status, RespHeaders} ->
             {ok, {{http_version, Status, aws_lib:status_text(Status)}, RespHeaders, <<>>}, Config1};
         {response, nofin, Status, RespHeaders} ->
-            {ok, Body} = gun:await_body(ConnPid, StreamRef, ?DEFAULT_HTTP_TIMEOUT),
-            {ok, {{http_version, Status, aws_lib:status_text(Status)}, RespHeaders, Body}, Config1};
+            case gun:await_body(ConnPid, StreamRef, ?DEFAULT_HTTP_TIMEOUT) of
+                {ok, Body} ->
+                    {ok, {{http_version, Status, aws_lib:status_text(Status)}, RespHeaders, Body},
+                        Config1};
+                {error, Reason} ->
+                    {error, Reason}
+            end;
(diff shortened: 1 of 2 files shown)


the-mikedavis (approved):
Looks good 👍

This might be an opportunity for `maybe` with all of the nested `case`s, but I don't feel too strongly about that

PR #109, opened 2026-07-01
Require a configured assume_role for auth validation; never use the instance role

sdewhitt:
[This commit](https://github.com/amazon-mq/rabbitmq-aws/pull/68/changes/120e47dcf9ca0d295556adf837c95761d7b65504) enforced auth validation requests to use the assume_role ARN passed into the broker's config when handling aws credentials. However, `resolve_request_state/1` built the request's aws_lib state from a bare `aws_lib:new()` whenever no `aws.arns.assume_role_arn` was configured, which falls back to the broker's ambient EC2 instance credentials. An instance role can be far more privileged than the role a customer would attach to their own secret/bucket, so a validate request could resolve ARNs the caller's intended role never could.

**Approach:** auth validation resolves ARNs **only** under the assume_role configured in the broker config (`aws.arns.assume_role_arn`), and refuses when none is set. The broker's ambient EC2 instance role is never used, and there is no configuration path to re-enable that fallback. With no role configured, the request is refused with `config_conflict` (422) before any secret fetch or outbound connection, preserving the endpoint's ARN-first ordering. Configured assume_role behavior is unchanged.

> Note: an earlier revision of this branch gated the fallback behind an opt-in `aws.auth_validation.allow_instance_role_fallback` flag. That was superseded -- per review we decided the endpoint should never use the instance role, so the flag was removed and the no-role case now always rejects.

  - `src/aws_auth_validate_ldap.erl`: the no-role branch of `resolve_request_state/1` returns `config_conflict` (422) via a fixed-category reason constant (no LDAP/secret detail); configured-role resolution and the STS-failure path are unchanged.
  - `priv/schema/aws.schema`: no new mapping (the interim flag mapping was removed).
  - `test/aws_auth_validate_tests.erl`: the not-configured case asserts a flat `config_conflict` refusal and that neither `assume_role` nor `resolve_arn` is called; the bind-raise and cacert-resolution fixtures configure an assume_role and stub `aws_iam:assume_role` to reach the resolve/bind paths they exercise.

### DIFF
--- a/.github/workflows/build-rabbitmq-server.yaml
+++ b/.github/workflows/build-rabbitmq-server.yaml
@@ -28,7 +28,7 @@ jobs:
             otp-version: 26
             elixir-version: 1.16
     steps:
-      - uses: erlef/setup-beam@fc68ffb90438ef2936bbb3251622353b3dcb2f93 # v1.24.0
+      - uses: erlef/setup-beam@54075bcc5e249e4758d363f27d099f55d843f124 # v1.24.1
         with:
           otp-version: ${{ matrix.otp-version }}
           elixir-version: ${{ matrix.elixir-version }}
@@ -37,7 +37,7 @@ jobs:
           git clone --branch ${{ matrix.rmq-version }} https://github.com/rabbitmq/rabbitmq-server.git ${{ github.workspace }}/rabbitmq-server
       - id: build-rabbitmq-server
         run: make -C ${{ github.workspace }}/rabbitmq-server
-      - uses: actions/cache/save@2c8a9bd7457de244a408f35966fab2fb45fda9c8 # v6.0.0
+      - uses: actions/cache/save@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
         with:
           path: ${{ github.workspace }}/rabbitmq-server
           key: rmq-${{ matrix.rmq-version }}-otp-${{ matrix.otp-version }}-elixir-${{ matrix.elixir-version }}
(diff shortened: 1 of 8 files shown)


the-mikedavis (approved):
Looks clean to me 👍

lukebakken (changes requested):
Reviewed the full diff plus the surrounding `validate/1` pipeline. Ran eunit (140/140), xref (baseline unchanged), and dialyzer (no new warning). The core change is correct: the no-role refusal is ordered before any assume_role, secret fetch, or outbound connection; `config_conflict` maps to 422; and the tests are non-vacuous (the not-configured case hard-errors if resolve_arn is reached and asserts zero assume_role/resolve_arn calls). Nice work.

Three items, all in comments/semantics rather than logic.

## M1 (please fix before merge): the `resolve_request_state/1` docstring contradicts itself

The new mandatory-role paragraph was appended without removing the sentence it invalidates. Lines 206-207 still say:

> With no role configured, a default state is used and ARNs resolve with the ambient role -- the prior behavior.

That is exactly the behavior this PR removes, and it contradicts lines 214-223 ("We therefore never use the instance role ... refuse with config_conflict"). On a security fix, a comment stating the opposite of the code is a real maintenance hazard. Rework the earlier paragraph (199-207) to describe the new invariant.

## L1: second stale comment at `resolve_arn/2` (lines 1118-1119)

Still describes the state as "a default state (broker's ambient role) or one with the operator-configured assume_role assumed." The default-state branch no longer exists; it is always the assumed role now. Folds in with the M1 edit.

## L2 (note, not a blocker): `config_conflict` reuse is a slight semantic stretch

The existing `config_conflict` flags a conflict between two request fields (use_ssl + use_starttls). Here it flags a missing broker-side prerequisite (aws.arns.assume_role_arn unset), driven by operator config rather than the caller's request. Reuse is defensible under the deliberate "no new category" constraint and the reason string is clear and actionable, so I would not add a category. Just noting the semantics differ.

PR #115, opened 2026-07-06
Parse URIs through one `uri_string`-based function

lukebakken:
Fixes #100.

`aws_lib:parse_uri/1` was a hand-rolled URI parser that split on `://` and `/`. A scheme-less, relative, or empty input matched neither case clause and crashed with `case_clause`. It also duplicated work that `aws_lib_uri:parse/1` already did via the RFC 3986 compliant `uri_string:parse/1`, and the plugin carried two URI representations (a bespoke `#uri{}` record plus the triple that adapter produced).

## What changed

`aws_lib_uri` is now a thin, opaque wrapper over the OTP `uri_string` module -- the single URI parser for the plugin:

- `parse/1` returns an opaque `uri()` (backed by `uri_string:uri_map/1`) or `{error, {malformed_uri, _}}` for input with no host or a `uri_string` parse error, rather than crashing
- callers read components only through accessors: `host/1`, `port/1` (scheme-defaulted), `path/1`, `query/1`, `target/1`
- `target/1` assembles the request line as path plus the raw query, byte-for-byte what was parsed -- what the Gun request target and the SigV4 signature must agree on
- removed `aws_lib:parse_uri/1` and its helpers, the `#uri{}` record and the types that only supported it, `aws_lib_uri:build/1` (its one caller now composes the metadata URL with `uri_string:recompose/1` directly), and `parse_userinfo/1,2`
- renamed `build_query_string/1` to `compose_query/1` to mirror `uri_string`

The opaque type structurally enforces the single-parser rule: dialyzer rejects any caller that reaches past the accessors.

## Behavior note

`instance_metadata_url("")` now yields `http://<host>` instead of `http://<host>/` (no trailing slash), at the one call site that discards the path and uses only host and port. No other call-site behavior changes.

## Testing

- reworked the `aws_lib_uri` eunit tests to the accessor API and added `target/1` coverage
- added a `proper`-based property suite (4 properties, 1000 iterations each): parsing never crashes on arbitrary input, well-formed URIs round-trip through the accessors, `path/1` always starts with `/`, and `target/1` preserves the query
- full eunit suite and the property suite pass; dialyzer is clean for the changed modules with no opacity violations


sdewhitt (approved):
Looks good to me! A bit of a nitpick, but i didn't see any comments specific to why we added `proper` as a new TEST_DEPS, but I looked through its uses and it doesn't seem like a blocking issue, especially if testing dependencies are not scrutinized the same as normal dependencies

src/aws_lib_sign.erl
```diff
@@ -31,7 +31,7 @@ headers(Request, undefined) ->
 headers(Request, PayloadHash) ->
     RequestTimestamp = local_time(),
     URI = aws_lib_uri:parse(Request#request.uri),
```
the-mikedavis: It's a more invasive change but I'd prefer to return a Result-like type (`{ok, GoodValue} | {error, Reason}`) since it makes it easy to swallow errors like this.

I believe the `host/1` call below would function_clause, where it would be better to have a badmatch on this. Or we could do this Elixir-style and either `throw/1` or `error/1` if an URI is invalid.

lukebakken: I originally wanted this then the 🧞 said "too much change" and I agreed 😸

the-mikedavis (approved):
It's a rougher diff but I like having the clean typing. And of course property tests 🚀

PR #116, opened 2026-07-07
Fix: Fail closed on the no-ARN branch instead of a default aws_state

sdewhitt:
The HTTP validation backend handed the no-ARN branch a usable aws_lib:new() state. That state carries undefined credentials, and aws_lib's credential chain (ensure_credentials_valid/1 -> refresh_credentials/1) falls back to env/file/EC2 instance metadata if any ARN fetch were ever attempted under it -- which would let a customer request resolve an ARN under the broker's ambient, potentially over-privileged instance role (a confused-deputy breach). Today that path is never taken, but only by an invariant spread across request_references_arn/1 and build_client_ssl_opts/1 agreeing on the same ARN-key set; a future ARN-backed field or an unconditional resolve would silently route to IMDS.

Store the sentinel `none' on the no-ARN branch and refuse it in resolve_arn/2, so the instance role is unreachable by construction rather than by a cross-function invariant. build_client_ssl_opts/1 defaults a missing aws_state to `none' too, preserving the fail-closed contract. The credential-free reachability probe is unaffected: it resolves zero ARNs, so it never reaches resolve_arn/2. The `none' clause maps to the existing fixed input_invalid/ARN-resolve category, leaking nothing.

Add regression tests: a no-ARN probe succeeds resolving nothing, a build_client_ssl_opts/1 with `none' + a cacertfile_arn fails closed without calling the resolver, and an ARN request with no configured assume_role returns config_conflict before any fetch.

### DIFF
--- a/src/aws_auth_validate_http.erl
+++ b/src/aws_auth_validate_http.erl
@@ -273,7 +273,21 @@ validate(Body) when is_map(Body) ->
 resolve_request_state(Params) ->
     case request_references_arn(Params) of
         false ->
-            {ok, Params#{aws_state => aws_lib:new()}};
+            %% No ARN is referenced, so this request resolves nothing and needs
+            %% no credentials. Do NOT hand out a usable aws_lib:new() state here:
+            %% that state carries undefined credentials, and aws_lib's credential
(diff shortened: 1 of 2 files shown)


the-mikedavis (approved):
💪

PR #118, opened 2026-07-07
Feature: OAuth Validation Endpoint

sdewhitt:
*Issue #43*

Allow customers to conveniently validate their OAuth 2.0 broker configs via /api/aws/auth/validate/oauth with more descriptive (but safe) error messages instead of blindly guessing error causes, changing the config, and restarting the broker (~30 min per restart).

This endpoint extends off of the validation endpoint introduced in PR #55, with this PR focusing on OAuth. The shared framework (Cowboy handler, registry, concurrency semaphore, backend behaviour, config schema, audit logging) is reused unchanged; this PR adds the OAuth backend, its https-only SSRF address policy, OIDC discovery, and JWKS well-formedness checking.

**Scope: reachability + JWKS well-formedness.** Without a real access token we cannot assert end-to-end token verification (signature/aud/scope against a live JWT). What we *can* confirm is that the signing-key source the broker would use is reachable, TLS-valid, and serves a well-formed JWKS. So an `ok` (204) means every configured HTTPS endpoint resolved and connected over TLS, and the JWKS endpoint returned HTTP 200 with a JSON object containing a non-empty `"keys"` array. It does **not** mean any particular token will validate -- the same caveat the LDAP `dn_lookup_base` and HTTP reachability checks carry, and it must be reflected in customer docs so the result is not over-trusted. A credentialed mode (accept a real token + assert full decode/verify against the fetched keys with `verify_aud`/`resource_server_id`) is intentionally deferred.

### Core subsystem (src/aws_auth_validate_oauth)
  - aws_auth_validate_oauth -- the OAuth backend (aws_auth_validate_backend behaviour: method_name/0, validate/1, allowed_fields/0). Pipeline: pure input validation (URL/resource_server_id/ssl_options shape) → SSRF address guard → assume-role + ARN resolution (cacertfile/certfile/keyfile) → ephemeral httpc probe over (m)TLS → optional OIDC discovery → JWKS fetch + well-formedness check → fixed-category result.
  - **OIDC discovery:** when only `issuer` is given, the backend fetches `{issuer}/.well-known/openid-configuration`, JSON-decodes it, and extracts `jwks_uri`; the derived URL is re-run through the SSRF guard and resolve-and-pin before it is fetched. A discovery doc that is unreachable or lacks `jwks_uri` maps to `auth_failed`.
  - **JWKS well-formedness:** success requires HTTP 200 AND a body that decodes to a JSON object whose `"keys"` is a non-empty array. A reachable endpoint that returns a non-200, an empty `"keys":[]`, or a non-JWKS body is `auth_failed` (reachable but not serving a usable signing-key source) -- this closes the false-pass where `jwks_uri` points at some other 200-returning service.
  - URL parsing rejects non-https schemes, out-of-range ports, userinfo (`user:pass@host`), and pre-existing query strings (OIDC discovery appends its own well-known path).
  - Accepted request fields mirror the customer's `oauth2.*` config so a config can be pasted as-is: `jwks_uri` and/or `issuer` (at least one required), `resource_server_id` (optional, non-empty-string check only -- not network-probed), and `ssl_options` (`verify`, `depth`, `versions`, `sni`, `cacertfile_arn`, `certfile_arn`, `keyfile_arn`).
  - Mutual TLS: `certfile_arn` (S3-hosted client cert PEM) + `keyfile_arn` (Secrets Manager client key PEM) are resolved and decoded into in-memory ssl `{cert,_}`/`{key,_}` options. `verify_peer` with no obtainable trust anchor fails loudly (`tls_failed`) rather than silently downgrading to `verify_none`.

### SSRF address policy (the central security concern, as with HTTP)
  Like HTTP, the validator issues requests to customer-supplied URLs from the broker's network position, so SSRF is front-and-center. Two-phase guard:
  - **Pure phase:** scheme allowlist (**https only** -- stricter than HTTP's http+https, since signing keys must travel over TLS) + literal-IP classification against a broker-infra denylist, before any DNS or network.
  - **Network phase:** resolve the host to all A/AAAA records, deny if any is infra, then connect to the **pinned IP literal** (original hostname kept only in the Host header + SNI) so httpc cannot re-resolve to a different address -- the DNS-rebinding/TOCTOU defense. This also covers the `issuer`-derived `jwks_uri`, which is guarded and pinned before it is fetched.
  - Denylist is **broker-infra only** (IMDS 169.254.169.254 + fd00:ec2::254, loopback, link-local, unspecified), with v6-embedded-v4 unwrapping (mapped/compatible/NAT64/6to4) so a v6 encoding cannot smuggle a denied v4. RFC1918 / unique-local are **deliberately allowed** -- a customer's IdP may live in their VPC. Redirects are not followed.

### Config (priv/schema/aws.schema, aws_sup, registry)
  - No new config schema or supervised workers: the OAuth method reuses the existing `aws.auth_validation.*` settings and the concurrency semaphore.
  - The `oauth` method is **opt-in** in the registry (defaults disabled even when `ldap` is enabled), so enabling validation for LDAP does not silently bring the OAuth probe online -- same posture as `http`.

### Tests
  - aws_auth_validate_oauth_tests (eunit, 45 tests): behaviour callbacks; pure input validation (missing-both jwks_uri/issuer, non-https scheme, URL shape -- userinfo/query/port, resource_server_id, every ssl_options fixed reason including the mTLS cert-without-key pairing); SSRF classify_ip (IMDS/loopback/link-local denied, v6-encoded IMDS denied) and the loopback-relax flag; mocked-network JWKS well-formedness (valid → ok, empty `keys` / non-JWKS body / non-200 → auth_failed, transport error → connection_failed); mocked OIDC discovery (success, missing jwks_uri → auth_failed); the assume_role guardrail (an ARN-referencing request with no configured role → config_conflict with resolve never called; with a role → resolves and succeeds; a no-ARN reachability request needs no role); and R6 no-leak (a resolved secret never appears in a rendered error).
  - aws_auth_validate_oauth_SUITE (Common Test, 10 cases): an in-process inets stub HTTPS server (self-signed and CA-signed listeners) exercises the real httpc probe end to end -- direct-jwks_uri success, issuer → OIDC discovery → derived jwks_uri success, empty/non-JWKS/non-200 JWKS bodies → auth_failed, discovery doc lacking jwks_uri → auth_failed, closed port → connection_failed, self-signed under `verify_peer` → tls_failed, the `verify_peer` + customer-CA (`cacertfile_arn`) success path (regression guard that the CA is passed to ssl as raw DER, and end-to-end proof of the assume_role → ARN-resolve path), and a response-contract guard that no fixed reason leaks the target host/port/URL. Skips gracefully if openssl or the local stub is unavailable.
  - aws_auth_validate_mgmt_SUITE: oauth dispatch cases -- disabled-by-default (opt-in) → 404, enabled + mocked dispatch → 204, no-secret-in-response.
  - Internal helpers are exposed under `-ifdef(TEST)` for direct unit coverage, mirroring the LDAP/HTTP backends.

### Refactors included in this PR
  Two follow-on changes ride with the OAuth backend; both are behaviour-preserving (all existing tests unchanged) and land as their own commits for review.

  1. **OAuth catch-up to the LDAP/HTTP parity fixes.** The OAuth backend was authored before several security fixes landed in the LDAP/HTTP backends and had drifted behind. Ported: the assume_role guardrail (resolving any ssl_options ARN requires a configured `aws.arns.assume_role_arn`; refuse with `config_conflict` rather than fall back to the broker EC2 instance role); the httpc profile-scan fix (claim a pool slot only by starting it and advance past `already_started`/`already_present` instead of stealing a peer's in-use profile); the cacerts-DER fix (pass raw `'Certificate'` DER to ssl, not `pem_entry_decode` records, which ssl silently ignores → unknown_ca → spurious tls_failed); and the loopback SSRF toggle (`auth_validation_allow_private_networks`, relaxing only loopback for local integration tests -- IMDS/link-local stay denied).

  2. **Shared-module extraction (ssl / net / httpc).** The three backends carried ~1000 lines of byte-identical code, so every security fix had to be applied three times or drift (which is how OAuth fell behind in the first place). Extracted into three modules: `aws_auth_validate_ssl` (ARN resolution + assume-role guardrail, ssl_options validation, PEM decoders, verify-default policy, TLS-error classification, bounded timeout), `aws_auth_validate_net` (the SSRF/DNS-rebinding guard, parameterized by a policy map), and `aws_auth_validate_httpc` (the ephemeral profile pool). Backend surface differences are absorbed by options maps; each backend keeps its own fixed reason strings. **The LDAP backend deliberately does not share `aws_auth_validate_net`** -- its address policy is stricter (it denies all private/RFC1918/CGNAT ranges, not just broker infra), so it keeps its own classifier. Net effect: http 1277 → 748 lines, oauth 1073 → 593, ldap 1142 → 1058. A new `aws_auth_validate_shared_tests` (9 tests) locks in cross-backend equivalence (http/oauth classify_ip parity across the v4/v6 matrix, loopback-flag parity, resolve_arn fail-closed) and pins the *intentional* http-vs-oauth scheme divergence so a future refactor cannot accidentally unify it.

### Response contract
  Reuses the existing fixed category set -- **no new category** is introduced. All failures return `{"error": "<category>", "message": "<constant>"}`, never a raw httpc error, target URL, response body, or resolved secret.
  OAuth-relevant categories: input_invalid / connection_failed / tls_failed (400), auth_failed (422, borrowed for "reachable but not a usable JWKS/OIDC document"), config_conflict (422, for an ARN-referencing request with no configured assume_role), plus the inherited pipeline-level categories. 204 on success.

### Security notes
  - Method disabled by default and opt-in even when other methods are enabled; admin-tag gated.
  - Ephemeral connections only, on a dedicated per-validation httpc profile (no pooled/reused TLS sessions, so one validation's client identity cannot leak into the next). No broker/auth-backend state mutation.
  - ARNs (cacertfile/certfile/keyfile) are resolved only after all pure input validation passes, so malformed requests never trigger a secret fetch; resolution requires a configured assume_role and never uses the broker instance role.
  - `assume_role` (request-supplied) is deliberately not accepted (confused-deputy on the broker's global AWS credentials), consistent with #55 and #66.
  - https-only: plaintext signing-key sources are rejected up front.

### Excluded from this PR
  - **Credentialed / full token verification** (accept a real access token and assert signature/aud/scope against the fetched keys) -- needs a real test token and a definition of "expected outcome"; deferred.
  - **CGNAT (100.64.0.0/10) blocking** -- like the HTTP backend, OAuth blocks only broker-owned infra and allows customer-VPC ranges (a customer IdP may live there), so CGNAT is intentionally allowed. This follows the same denylist policy as #66; the LDAP backend's stricter all-private-ranges block is a separate, deliberate policy.


the-mikedavis (approved):
I pushed up a small regression test addition but otherwise I have no notes. Nice one! 🚀

lukebakken (approved):
Approving.

Reviewed all 12 files: the three extracted shared modules (`aws_auth_validate_ssl` / `_net` / `_httpc`), the new OAuth backend, the registry/schema wiring, and both refactored backends against the merge base.

The shared-module extraction is a faithful, near-mechanical lift of previously byte-identical code; LDAP and HTTP now delegate, with the per-backend R4 reason strings kept local so wording and existing tests are unchanged. The OAuth backend mirrors the reviewed HTTP backend, and the security-critical paths hold up:

- SSRF/DNS-rebinding: https-only allowlist (stricter than HTTP), v6-embedded-v4 unwrapping, resolve-and-pin, `autoredirect=false`, and the issuer-derived `jwks_uri` re-guarded and pinned before fetch. `oidc_discovery_jwks_uri_ssrf_denied_test_` now nails that last path shut (valid JWKS served at the IMDS target, so a pass can only come from the guard firing; asserts a single `httpc:request`, never the denied fetch)
- assume_role guardrail: the `none` credential sentinel makes the instance role unreachable by construction on the no-ARN branch; tested config_conflict / resolves-with-role / no-ARN-needs-none
- R6: network section wrapped to a fixed category; no-secret-leak asserted in unit, CT, and mgmt suites
- opt-in: `oauth` in `?OPT_IN_METHODS`, 404-by-default verified

Verified locally: eunit 464/464; dialyzer at the pre-existing baseline with zero warnings in any of the four new modules; xref clean; `aws_auth_validate_oauth_SUITE` 10/10 (incl. `custom_ca_verify_peer_returns_ok`, exercising assume_role + the cacerts-as-DER path end to end, and `self_signed_verify_peer_returns_tls_failed`); `aws_auth_validate_mgmt_SUITE` oauth group green (404 / 204 / no-secret) with all pre-existing groups unaffected.

PR #125, opened 2026-07-08
Decode XML attributes and document the parser contract

lukebakken:
Fixes #101.

`aws_lib_xml:parse/1` only read each element's `name` and `content`, so any data carried in XML attributes was silently dropped, and the mixed-content and repeated-sibling behaviours were undocumented. `maybe_decode_body/2` routes every `*/xml` response through this parser, so a caller consuming a richer response would get incomplete data with no error.

## What changed

- Attributes are decoded into a reserved `'@attributes'` key at the head of an element's value proplist. When an element has attributes AND only text, the text moves under a reserved `'#text'` key so it can sit beside them.
- Elements WITHOUT attributes decode exactly as before (byte-identical output), so the `DescribeVolumes` and STS `AssumeRole` consumers are unchanged.
- Repeated same-named siblings are preserved as repeated proplist keys (already the behaviour); this is now documented as part of the contract.
- The module now documents its full contract, including the scoped mixed-content and whitespace limitations (this is not a general-purpose XML decoder; it targets the AWS response shapes the plugin consumes).

## Testing

- example cases for attributes (text element, element with children, attribute-only) and repeated siblings
- **new end-to-end consumer tests** feeding real `DescribeVolumes` and `AssumeRole` XML through `parse/1` into `parse_volumes_response/1` and `parse_assume_role_response/2` (the latter in a new `aws_iam_tests`; `aws_iam` gains a `-ifdef(TEST)` export block). These previously had no coverage and guard the production shapes against future parser changes.
- **new proper-based round-trip suite** (`prop_aws_lib_xml_SUITE`) over generated element trees. Coverage of a 500-tree sample: 91.8% carry attributes, 40.4% have repeated siblings, avg depth 3.1 / avg 15 nodes. The generator draws element/attribute names and values from small fixed pools (xmerl_scan interns them as atoms, so unbounded names would exhaust the atom table) and bounds tree depth and breadth.

Full eunit and the property suite pass; the property is non-vacuous (it fails against the old attribute-dropping parser).

> Note: while running the full suite I found a pre-existing flaky test on `main` unrelated to this change (`aws_lib_config_tests` "imdsv2 is available" hits real IMDS). Filed separately as #124.

### DIFF
--- a/test/prop_aws_lib_xml_SUITE.erl
+++ b/test/prop_aws_lib_xml_SUITE.erl
@@ -1,0 +1,170 @@
+%% Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+%% SPDX-License-Identifier: Apache-2.0
+%% vim:ft=erlang:
+%% -*- mode: erlang; -*-
+
+%% Property-based tests for aws_lib_xml:parse/1. Complements the example-based
+%% cases in aws_lib_xml_tests. A generated element tree is rendered to an XML
+%% string, parsed back, and checked to round-trip: element names, attributes,
+%% text, nesting, and repeated siblings are all recovered.
+%%
+%% The generator produces the well-formed subset the parser targets (see the
+%% aws_lib_xml module doc): each element has EITHER text OR child elements, never
+%% mixed content, and names/values avoid whitespace and XML metacharacters so no
+%% escaping or whitespace-stripping is exercised. Those are the parser's
+%% documented lossy behaviors and are covered by example tests, not here.
+-module(prop_aws_lib_xml_SUITE).
+
+-compile(nowarn_export_all).
+-compile(export_all).
+
+-include_lib("proper/include/proper.hrl").
+-include_lib("common_test/include/ct.hrl").
+
+%% 200 iterations is ample: names, values, and attribute names are drawn from
+%% small fixed pools (see name/0, value/0), so the input space this explores is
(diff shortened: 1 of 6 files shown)


test/prop_aws_lib_xml_SUITE.erl
```diff
@@ -0,0 +1,170 @@
...
+        "Foo",
+        "Bar",
+        "region",
+        "id",
+        "status",
+        "value"
+    ]).
+
+%% A non-empty value drawn from a small FIXED pool. Like element names,
+%% xmerl_scan interns attribute values as atoms, so unbounded distinct values
```
the-mikedavis: I think that xmerl_scan makes atoms for names but not values:

```erlang
{ok, _} = application:ensure_all_started(xmerl),
{Element, _} = xmerl_scan:string("<x a=\"myvalue\"/>"),
io:format("~p~n", [Element]).
```

```erlang
{xmlElement,x,x,[],
            {xmlNamespace,[],[]},
            [],1,
            [{xmlAttribute,a,[],[],[],[{x,1}],1,[],"myvalue",false}],
            [],[],"/path/to/source",undeclared}
```

the-mikedavis (approved):
Super minor doc thing but otherwise lgtm

sdewhitt (approved):

the-mikedavis (approved):

PR #126, opened 2026-07-08
Decode all AWS JSON content types, not just 1.0 and plain

lukebakken:
Fixes #99.

`maybe_decode_body/2` only matched `application/x-amz-json-1.0` and `application/json`, so an `application/x-amz-json-1.1` response fell through to the catch-all and was returned undecoded, despite the documented contract that JSON responses are decoded automatically. Secrets Manager uses the JSON 1.1 protocol, so this would bite once `aws_sms` relies on the library's auto-decoding (issue #53).

## What changed

Decode any JSON subtype via a guard. `is_json_subtype/1` matches:

- plain `json`
- every AWS JSON protocol version (`x-amz-json-1.0`, `x-amz-json-1.1`, and any future `x-amz-json-*`)
- the RFC 6839 structured `+json` suffix

XML and unknown subtypes are unchanged (routed through `maybe_decode_xml/2`).

## Testing

- new positive cases: `x-amz-json-1.1` (the bug) and an `application/vnd.api+json` structured suffix
- new negative case: `application/octet-stream` must fall through undecoded (guards against the JSON match being too greedy)
- existing `x-amz-json-1.0`, `application/json`, `text/xml`, and `text/html` cases retained

Full eunit and dialyzer are clean for the changed module; the new tests are non-vacuous (they fail against the previous clause-based matching).

### DIFF
--- a/src/aws_lib.erl
+++ b/src/aws_lib.erl
@@ -578,14 +578,30 @@ local_time() ->
 %% @end
 maybe_decode_body(_, <<>>) ->
     <<>>;
-maybe_decode_body({"application", "x-amz-json-1.0"}, Body) ->
-    aws_lib_json:decode(Body);
-maybe_decode_body({"application", "json"}, Body) ->
-    aws_lib_json:decode(Body);
-maybe_decode_body({_, "xml"}, Body) ->
+maybe_decode_body({"application", Subtype}, Body) ->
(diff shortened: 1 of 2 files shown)


the-mikedavis (approved):

sdewhitt (approved):
looks good to me!

PR #135, opened 2026-07-09
Feature: TLS material-validation endpoint

sdewhitt:
*Issue #43*

## Summary

Customers configuring SSL certificate authentication or mTLS on their Amazon MQ for RabbitMQ brokers follow these tutorials ([SSL cert](https://docs.aws.amazon.com/amazon-mq/latest/developer-guide/rabbitmq-ssl-tutorial.html), [mTLS](https://docs.aws.amazon.com/amazon-mq/latest/developer-guide/rabbitmq-mtls-tutorial.html)). Both reference a CA bundle by ARN (`aws.arns.ssl_options.cacertfile` / `aws.arns.management.ssl.cacertfile`). Today a wrong ARN or a malformed/expired CA only surfaces at broker boot, so the customer discovers it after building a keystore and running the verification script.

This adds a `tls` validation method that checks that broker-side material up front, without a restart.

## What it does

`PUT /api/aws/auth/validate/tls` with a `target` (`listener` or `management`) and `ssl_options`:

- resolves the `cacertfile_arn`,
- decodes the PEM and confirms it holds at least one well-formed CA certificate,
- checks that no certificate is expired or not yet valid,
- validates the shapes of `verify`, `fail_if_no_peer_cert`, `depth`, `versions`.

Results reuse the existing categories: `204` on success, `400 input_invalid` for bad input / an unresolvable ARN / an unparseable CA, `400 tls_failed` for an expired or not-yet-valid certificate, and `422 config_conflict` when a `cacertfile_arn` is given but no `assume_role_arn` is configured.

For the mTLS tutorial (a CA on both the AMQP and management listeners), the customer validates each with a separate call (`target: listener`, then `target: management`).

## Scope

This validates the broker-side material only; it does not perform a live handshake. A pass means the CA material is usable, not that TLS/mTLS as a whole is working. It does not cover client-certificate chaining, the common-name to user mapping, network reachability, or whether the broker is currently running this configuration. The tutorial scripts (`ssl.sh` / `mtls.sh`) remain the end-to-end check; this is a pre-flight that rules out the common CA/ARN/shape errors first.

## Notes

- The method is opt-in (disabled unless explicitly enabled), consistent with the other non-default methods.
- Unlike the ldap/http/oauth methods, this makes no outbound connection; the only network call is the ARN resolution.

## Testing

- Unit tests (`aws_auth_validate_tls_tests`): input validation, the assume_role guardrail, ordering, no-leak of resolved material, and certificate-validity logic.
- Integration tests (`aws_auth_validate_tls_SUITE`): the full HTTP pipeline against a broker node, with the ARN/assume-role boundaries mocked.

### DIFF
--- a/src/aws_auth_validate_http.erl
+++ b/src/aws_auth_validate_http.erl
@@ -167,13 +167,12 @@
 -define(AUTH_RESPONSE_CODES, [200, 201]).
 
 %% SSRF guard toggles (see the SSRF guard section below for the full rationale).
-%% The address policy (classify_ip/1 + the probe-layer resolve+pin) is now live,
-%% so ?ENFORCE_ADDRESS_POLICY is TRUE. It is retained as a defined marker of that
+%% The address policy (classify_ip/1 + the probe-layer resolve+pin) is live, so
+%% ?ENFORCE_ADDRESS_POLICY is TRUE. It is retained as a defined marker of that
 %% state; enforcement itself runs unconditionally via url_allowed/1 and the
-%% probe-layer resolve+classify.
-%% NOTE: AppSec review + pen test are still pending (parent FAQ 2.1); the method
-%% remains opt-in (aws_auth_validate_registry ?OPT_IN_METHODS) so it is not live
-%% until an operator explicitly enables it.
+%% probe-layer resolve+classify. The method is opt-in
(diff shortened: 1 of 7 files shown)


the-mikedavis (approved):
No blockers from me, looks good 👍

lukebakken (approved):
LGTM. CI green.

PR #139, opened 2026-07-10
Enhancement: OAuth Customer-Supplied Token Verification

sdewhitt:
*Issue #144*

## Summary

The OAuth validation endpoint introduced in #118 proves **reachability + JWKS well-formedness**: a `204` means the signing-key source the broker would use is reachable, TLS-valid, and serves a well-formed JWKS -- not that a token actually verifies against it. This PR closes that gap with an **optional, opt-in** capability: **customer-supplied token verification**. The operator supplies their own access token (minted out of band) and the broker verifies its signature, expiry, and audience against the JWKS it just fetched -- **with no secret transiting the endpoint**.

When `access_token` is absent, behaviour is identical to #118 (pure JWKS reachability) -- no new required fields, no credential handling.

## What it does

`PUT /api/aws/auth/validate/oauth` with `jwks_uri` (or `issuer`) plus, optionally, `access_token` and `resource_server_id`.

The operator mints a token out of band -- a single `curl` to their IdP's token endpoint -- and supplies it in the request; the broker verifies it against the JWKS it fetched. Because the token is minted externally, **no client secret ever transits this endpoint**.

- **Pure phase:** three-segment JWT shape check, base64url header decode, and an **asymmetric algorithm allowlist** (RS*/ES*/PS*). `none` and every HMAC (HS*) algorithm are refused before any network -- a JWKS holds only public keys, so an HS* token would otherwise be "verified" by HMAC-ing with the public key as the shared secret (the classic alg-confusion forgery).
- **After the JWKS fetch (no extra network):** select the key whose `kid` matches, verify the signature with `jose_jwt:verify_strict/3` (algorithm **pinned** to the header alg), check `exp`/`nbf` (same semantics as `rabbit_auth_backend_oauth2:validate_token_expiry/1`), and -- when `resource_server_id` is supplied -- require it in the token `aud`.

**Crypto parity without coupling.** Verification reuses the broker's own crypto library (`jose`) rather than the `rabbit_auth_backend_oauth2` backend, so the signature decision matches what the broker computes **without forcing the oauth2 backend onto every `aws`-plugin user** (LDAP/IAM/ARN-only installs). `jose`'s runtime deps (`kernel/stdlib/crypto/asn1/public_key`) are already present; it is added to `DEPS` only.

## Response contract

Reachability outcomes reuse the existing categories. Verification adds **two** new categories so an operator is informed beyond a flat `auth_failed`:

| Category | Status | Meaning |
|---|---|---|
| `token_invalid` | 422 | Bad signature or no matching JWKS key -- a **real config mismatch**; the broker would also reject live tokens. |
| `token_expired` | 422 | Expired (`exp`) or not-yet-valid (`nbf`) -- **transient**, just re-mint and retry. |

An audience mismatch stays `auth_failed` (422); a malformed or unsupported-alg token is `input_invalid` (400). These two categories are safe to distinguish (unlike the coarse reachability categories, which stay coarse to avoid SSRF reconnaissance) precisely because they describe the caller's *own* token -- not the broker's infrastructure or an SSRF target -- so they leak nothing R4 guards. `204` on success. No raw httpc error, target URL, response body, or token claim is ever echoed.

## Shared-module changes

**Wildcard-cert hostname check.** `apply_verify_default/2` now adds `{customize_hostname_check, [{match_fun, public_key:pkix_verify_hostname_match_fun(https)}]}` on every `verify_peer` path. Without it, ssl's default hostname check rejects a single-label wildcard cert covering a multi-label host -- e.g. every Cognito hosted-UI domain, where a JWKS/token host `foo.auth.<region>.amazoncognito.com` is served by a `*.auth.<region>.amazoncognito.com` cert -- which `curl`/`openssl`/browsers all accept under the RFC 6125 https rules. The symptom was a spurious `tls_failed` for perfectly valid IdP endpoints, the exact false-negative this endpoint exists to prevent. This does **not** relax verification (a genuine hostname mismatch still fails; verified live against a bad-SAN host) and does not override a caller-supplied `customize_hostname_check`; `verify_none` is untouched. Lives in the shared module so both the http and oauth backends get it (ldap uses a different TLS path).

## Scope

Verification asserts **signature + `exp`/`nbf` + `aud`**. It does not assert scope authorization -- that stays an over-trust caveat for the customer docs, same as the LDAP `dn_lookup_base` and HTTP reachability checks. It is a pre-flight that rules out the common signing-key/token/config errors before a broker restart.

## Security notes

- The capability is opt-in per request (absent `access_token` = unchanged reachability); the `oauth` method itself remains disabled-by-default and admin-tag gated.
- No secret is handled: the token is minted by the customer out of band. Any `ssl_options` ARN (cacert/cert/key) is still resolved only after pure validation, only under a configured assume_role, never with the broker instance role.
- Request-supplied `assume_role` remains deliberately unsupported (confused-deputy on the broker's global AWS credentials), consistent with #55/#66/#118.
- Ephemeral connections only, on a dedicated per-validation httpc profile; no broker/auth-backend state mutation.
- Alg-confusion is closed by the asymmetric-only allowlist plus `verify_strict/3` pinning.
- Wildcard-cert verification uses the standard https match rules but stays strict: a genuine cert/hostname mismatch still fails `verify_peer` (no verification is disabled).

## Testing

- **Unit (`aws_auth_validate_oauth_tests`, eunit -- 72/72):** pure-phase shape and alg-allowlist checks (`none`/HS256 rejected), JWK selection by `kid`, and signature/exp/nbf/aud verification (valid; wrong-key -> `token_invalid`; expired/nbf -> `token_expired`; aud mismatch -> `auth_failed`); end-to-end through `validate/1` with a mocked JWKS fetch, including an expired-token -> `token_expired` case and an R6 no-leak check that a verified token's claims never appear in the result.
- **Shared (`aws_auth_validate_shared_tests`, eunit -- 13/13):** cross-backend `apply_verify_default/2` cases, including the wildcard hostname-check (adds the https `match_fun` on explicit and defaulted `verify_peer`; leaves `verify_none` alone; does not override a caller-supplied `customize_hostname_check`).
- **Integration (`aws_auth_validate_oauth_SUITE`, Common Test -- 12/12):** JWKS reachability / OIDC discovery / well-formedness / TLS paths against an in-process stub HTTPS server.
- **Manual (against `RabbitMqOAuth2TestStack` / Cognito):** `access_token` verification returns `204` under **default `verify_peer`** against the wildcard-cert host (previously `tls_failed`); negative controls (tampered claim / bad signature -> `token_invalid`, not-a-JWT -> `input_invalid`, bad-SAN host -> `tls_failed`) all behave correctly.


src/aws_auth_validate_oauth.erl
```diff
@@ -412,6 +573,149 @@ do_oauth_validate(Params) ->
...
+%% non-numeric exp/nbf is treated as expired (the token is malformed w.r.t. time).
+check_token_expiry(Claims) ->
+    Now = os:system_time(seconds),
+    case check_exp(maps:get(<<"exp">>, Claims, undefined), Now) of
+        ok -> check_nbf(maps:get(<<"nbf">>, Claims, undefined), Now);
+        Err -> Err
+    end.
+
+check_exp(undefined, _Now) ->
+    ok;
```
the-mikedavis: We might need to check the server setting `auth_oauth2.require_exp` here - otherwise the validation might say that a token with no expiry is OK while it would not be accepted according to server settings. This option was introduced really recently, so this is mainly a future-proofing measure.

src/aws_auth_validate_oauth.erl
```diff
@@ -412,6 +573,149 @@ do_oauth_validate(Params) ->
...
+
+check_nbf(undefined, _Now) ->
+    ok;
+check_nbf(Nbf, Now) when is_number(Nbf) ->
+    case trunc(Nbf) > Now of
+        true -> {error, token_expired, ?REASON_TOKEN_EXPIRED};
+        false -> ok
+    end;
+check_nbf(_Nbf, _Now) ->
+    {error, token_expired, ?REASON_TOKEN_EXPIRED}.
```
the-mikedavis: `?REASON_TOKEN_EXPIRED` might be too broad here - if these values are not valid then an invalid reason might be more appropriate

src/aws_auth_validate_oauth.erl
sdewhitt: Sounds good, pushing a commit rn with these changes:
- Added require_exp/0 reading application:get_env(rabbitmq_auth_backend_oauth2, require_exp, false)
  - the same source the real backend uses.
- check_exp(undefined, ...) now rejects a no-exp token with token_invalid (?REASON_TOKEN_MISSING_EXP) when the server requires exp, otherwise still accepts it. This prevents the false-pass where the endpoint says a no-exp token is OK but the live broker would reject it.
- Tests: verify_token_no_exp_ok_by_default, verify_token_no_exp_rejected_when_required (→ token_invalid), verify_token_present_exp_ok_when_required.

src/aws_auth_validate_oauth.erl
sdewhitt: Just pushed a commit with these changes:
- The non-numeric exp/nbf fallthrough clauses now return token_invalid with a dedicated ?REASON_TOKEN_BAD_TIME_CLAIM ("non-numeric exp or nbf claim") instead of token_expired. A malformed time claim means the token is invalid, not merely expired. The genuinely expired/not-yet-valid numeric cases still correctly return token_expired.
- Tests: verify_token_non_numeric_exp_is_invalid, verify_token_non_numeric_nbf_is_invalid (both → token_invalid).

the-mikedavis (approved):

src/aws_auth_validate_oauth.erl
```diff
@@ -412,6 +578,169 @@ do_oauth_validate(Params) ->
...
+check_nbf(_Nbf, _Now) ->
+    %% present but non-numeric -> malformed claim, not "not yet valid"
+    {error, token_invalid, ?REASON_TOKEN_BAD_TIME_CLAIM}.
+
+%% When a resource_server_id was supplied, require it in the token `aud' (a
+%% string or a list of strings, per RFC 7519). When none was supplied, the aud
+%% is not asserted (signature + expiry only).
+check_token_audience(_Claims, undefined) ->
+    ok;
+check_token_audience(Claims, ResourceServerId) ->
```
lukebakken, line 767: **`verify_aud=false` is not honored.** `check_token_audience` enforces the audience whenever a `resource_server_id` is supplied. The broker gates this on `verify_aud` (default `true` but operator-settable to `false` in `rabbit_oauth2_resource_server`, and it then falls back to `find_unique_resource_server_without_verify_aud`). If an operator runs `auth_oauth2.verify_aud = false` and still passes `resource_server_id` to the probe, a token whose `aud` omits the id is accepted by the broker but rejected here -- a false fail. Deciding whether (and how) the probe should read `verify_aud` is a design call, so I left it for you.

src/aws_auth_validate_oauth.erl
```diff
@@ -412,6 +578,169 @@ do_oauth_validate(Params) ->
...
+%% Read the server's rabbit_auth_backend_oauth2 `require_exp' setting (recent
+%% option). Absent/false -> a token without exp is accepted; true -> it is
+%% rejected, matching what the live broker would do. Read from the oauth2
+%% backend's app env, the same source the backend itself uses.
+require_exp() ->
+    application:get_env(rabbitmq_auth_backend_oauth2, require_exp, false) =:= true.
+
+check_nbf(undefined, _Now) ->
+    ok;
+check_nbf(Nbf, Now) when is_number(Nbf) ->
```
lukebakken: **`nbf` is validated, but the broker never checks it.** There is no `nbf`/`not_before` handling anywhere in `rabbit_auth_backend_oauth2` (`validate_token_expiry/1` inspects only `exp`). A valid token with an `nbf` slightly in the future (clock skew, or a post-dated token) is accepted by the broker but returns `token_expired` here -- a false fail, made worse by zero leeway. The doc comment above `check_token_expiry` claiming "same semantics as rabbit_auth_backend_oauth2" is inaccurate for this path. Options: drop the `nbf` check for strict parity, or keep it as a deliberate extra safety net and correct the comment. Your call.

src/aws_auth_validate_oauth.erl
```diff
@@ -412,6 +578,169 @@ do_oauth_validate(Params) ->
...
+                        {error, _, _} = Err -> Err;
+                        ok -> check_token_audience(Claims, ResourceServerId)
+                    end
+            end
+    end.
+
+%% Select the JWKS entry matching the token's `kid'. When the token carries no
+%% `kid', accept a single-key JWKS (unambiguous); otherwise a missing kid on a
+%% multi-key set is unresolvable. A malformed (non-object) key entry is skipped.
+select_jwk(undefined, [Key]) when is_map(Key) ->
```
lukebakken: **No-kid token selection differs from the broker.** For a token with no `kid`, this accepts a single-key JWKS and verifies against that key. The broker resolves a no-kid token via the configured `default_key` (`uaa_jwt:decode_and_verify` -> `get_key_id/1` returns `undefined` -> `internal_oauth_provider.default_key`), not "the only key present". So a JWKS that serves one key carrying a `kid` plus a token with no `kid` passes here but the broker (with no `default_key` configured) rejects it -- a potential false pass. Whether to mirror the `default_key` behavior is a design decision.

src/aws_auth_validate_ssl.erl
```diff
@@ -151,11 +157,28 @@ resolve_request_state(Params, Opts) ->
...
+ssl_options_reference_arn(#{ssl_options := Map}, #{arn_keys := Keys}) ->
     lists:any(fun(K) -> maps:is_key(K, Map) end, Keys);
-request_references_arn(_Params, _Opts) ->
+ssl_options_reference_arn(_Params, _Opts) ->
+    false.
+
+%% A top-level param ARN counts as a reference only when the value is actually
+%% present (a non-empty binary): merely listing the key in param_arn_keys does
+%% not force an assume_role if the caller omitted the secret.
+params_reference_arn(Params, #{param_arn_keys := Keys}) when is_map(Params) ->
```
lukebakken: **`param_arn_keys` looks like dead infrastructure.** No backend's opts map sets `param_arn_keys` (http, ldap, and the oauth backend in this PR all omit it), so this `params_reference_arn/2` clause never matches and the assume_role guardrail for a top-level secret ARN it describes never fires. If this is intentional scaffolding for a future `client_secret_arn` change, a one-line note would help; otherwise it could be dropped until a caller actually sets it.

src/aws_auth_validate_oauth.erl
```diff
@@ -335,6 +418,79 @@ parse_ssl_options(Body, Acc) ->
...
+            end;
+        {error, _} ->
+            {error, input_invalid, ?REASON_TOKEN_MALFORMED}
+    end.
+
+%% Decode the first JWT segment (the protected header) into a JSON object map.
+%% Uses unpadded base64url (JWT compact serialization). Any decode/JSON failure
+%% is a malformed token.
+-spec token_header(binary()) -> {ok, map()} | {error, malformed}.
+token_header(Raw) ->
```
lukebakken, line 504: **Reuse opportunity (not a bug).** `token_header/1` hand-splits the compact JWT and JSON-decodes the header; `b64url_decode/1` (line 487) duplicates `jose_base64url:decode/1`; and `jwt_claims/1` (line 666) extracts claims via `element(2, JWT)` rather than the `#jose_jwt{fields = Fields}` record accessor. jose is now a direct dep, and the sibling `uaa_jwt_jwt` already uses `jose_jwt:peek_protected/1` for exactly this. Reusing jose here would remove a second JWT parser to keep in sync and avoid the fragile positional `element/2`. Worth considering; works correctly as-is.

lukebakken (commented):
## Review: config-parity of the OAuth token-verification path

The core value of this endpoint is catching OAuth config mismatches, so a divergence from how the live broker (`rabbit_auth_backend_oauth2`) actually verifies a token is a correctness issue: it produces a false pass (endpoint says OK, broker would reject) or a false fail. I checked every claim below against the broker source.

### Fixed on this branch (one commit each)

- 6184b37 **Default `require_exp` to true to match the broker** -- `require_exp/0` defaulted to `false`, but the broker defaults it to `true`, so a no-exp token was a false pass. Now defaults to `true` and mirrors the broker's coercion; the test that pinned the old default is rewritten
- 643a829 **Split a space-delimited string `aud` before matching** -- a binary `aud` was matched whole with `=:=`; the broker space-splits it first, so a multi-value string `aud` was a false fail. Now splits on spaces like `find_audience/2`
- 9cc6d74 **Reject a non-binary `kid` in the pure phase** -- a non-binary header `kid` crashed `select_jwk/2` and was misreported as `connection_failed`; now rejected cleanly as a malformed token

All oauth eunit tests pass (80), xref clean.

### Outstanding -- design calls left to you (see inline comments)

These are real divergences but the right resolution is a judgment call, so I did not auto-apply them:

1. **`verify_aud=false` not honored** -- aud is always enforced when a `resource_server_id` is supplied, ignoring the broker's operator-settable `verify_aud=false`
2. **`nbf` is checked but the broker never checks it** -- an over-strict divergence (false fail on clock skew); the parity comment is also inaccurate
3. **no-kid single-key selection** -- the broker resolves a no-kid token via the configured `default_key`, not "the only key in the JWKS"; potential false pass
4. **`param_arn_keys` is dead infrastructure** -- no backend sets it; flagging in case it is intentional scaffolding for a later `client_secret_arn` change
5. **jose helpers re-implemented** -- `token_header`/`b64url_decode`/`jwt_claims` duplicate jose facilities already used by the oauth2 backend

src/aws_auth_validate_oauth.erl
sdewhitt: Fixed in 3a86cec.
- Added verify_aud/0 (mirroring require_exp/0)
- check_token_audience/2 now skips the aud match when the broker's verify_aud is disabled
  - Token whose aud omits resource_server_id is no longer a false fail. 
- Added a test for the verify_aud=false skip and one pinning the default-enforced case.

src/aws_auth_validate_oauth.erl
sdewhitt: Fixed in 3a86cec
- Went with strict parity and dropped the nbf check entirely, since validate_token_expiry/1 inspects exp only
- Corrected the "same semantics" comment, which was the inaccurate part you flagged
- check_token_expiry/1 now checks exp alone

src/aws_auth_validate_oauth.erl
sdewhitt: Fixed in 3a86cec. Rather than model default_key, select_jwk/2 now rejects a no-kid token as token_invalid -- the conservative choice, since accepting the lone JWKS key was the false pass. Modeling default_key would couple this to the oauth2 provider config for a narrow case; rejecting matches the common "no default_key configured" broker outcome. Test inverted to expect the error.

src/aws_auth_validate_ssl.erl
sdewhitt: Yup, it was part of a different feature I scrapped. I just removed it in 559611c, and folded request_references_arn/2 back to its single ssl_options source and dropped the unused opts() field. No behavior change.

src/aws_auth_validate_oauth.erl
sdewhitt: Done in 3a86cec. token_header/1 now uses jose_jwt:peek_protected/1 + jose_jws:to_map/1 (the uaa_jwt_jwt idiom) and jwt_claims/1 uses the #jose_jwt{fields} accessor instead of positional element/2.

One gotcha worth flagging: peek_protected/1 lifts alg out into the #jose_jws{} record, so fields alone omits it -- jose_jws:to_map/1 is needed to get the full header back for the alg-allowlist check. Kept the 3-segment shape check and the alg/kid guards.

lukebakken (approved):
👍

PR #140, opened 2026-07-10
Classify AWS responses and stop retrying non-retriable errors

lukebakken:
Closes #80.

Stacked on #138 (`feature/expose-aws-error-details`); only the top commit is new here.

The retry loop in `api_request_with_retries` treated every failure the same: an access-denied or invalid-parameter 4xx burned all five retries with backoff before the caller saw the error, even though repeating the request could never succeed.

This PR classifies each failed response as `retriable` or `not_retriable` and returns non-retriable errors to the caller immediately:

- 5xx, HTTP 429, throttling error codes, and transport failures are retriable
- other 4xx client errors and unfollowed 3xx redirects are not

Throttling detection inspects the error code fields of the decoded body (`__type`/`code` for the JSON protocols, `Code` for XML error documents, per the Smithy protocol specs) and the `X-Amzn-Errortype` header rather than substring-matching the whole body, so a marker inside a message or resource name does not trigger a retry. The marker list collapses botocore's `_THROTTLED_ERROR_CODES` into case-insensitive substrings. An undecodable body falls back to a raw substring search.

Response interpretation moves into a new `aws_lib_response` module: `format_response/1`, `classify_response/2`, `maybe_decode_body/2`, `get_content_type/1`, `parse_content_type/1`, and `status_text/1` move out of `aws_lib`, which keeps the retry loop. `aws_lib_httpc` produces the raw response, `aws_lib_response` decides what it means, and a new `response_status/1` accessor reports the outcome without callers matching the raw tuple. `perform_request_reuse` formats the response once and passes the result to both consumers, so the body is decoded exactly once per response.

Verification: 541 eunit tests pass, `xref` clean, `dialyzer` clean in all touched files, `erlfmt` applied.

### DIFF
--- a/include/aws_lib.hrl
+++ b/include/aws_lib.hrl
@@ -163,10 +163,12 @@
     | {ok, {status_code(), body()}}
     | {error, term()}.
 
--type result_ok() :: {ok, {ResponseHeaders :: headers(), Response :: list()}}.
+%% The response payload is the decoded body: a proplist for JSON/XML content
+%% types, or the raw binary when the body could not be decoded.
+-type result_ok() :: {ok, {ResponseHeaders :: headers(), Response :: list() | binary()}}.
 -type result_error() ::
     {'error', Message :: reason_phrase(),
-        {ResponseHeaders :: headers(), Response :: list()} | undefined}
+        {ResponseHeaders :: headers(), Response :: list() | binary()} | undefined}
     | {'error', {credentials, Reason :: string()}}
     | {'error', string()}.
 -type result() :: result_ok() | result_error().
--- a/include/aws_lib.hrl
+++ b/include/aws_lib.hrl
@@ -158,11 +158,6 @@
 }).
 -type request() :: #request{}.
 
--type httpc_result() ::
-    {ok, {status_line(), headers(), body()}}
-    | {ok, {status_code(), body()}}
-    | {error, term()}.
-
 %% The response payload is the decoded body: a proplist for JSON/XML content
 %% types, or the raw binary when the body could not be decoded.
 -type result_ok() :: {ok, {ResponseHeaders :: headers(), Response :: list() | binary()}}.
--- a/src/aws_lib.erl
+++ b/src/aws_lib.erl
@@ -847,6 +847,14 @@ api_request_with_retries(
                             %% reuse mode the connection is handed back for the
                             %% next request; in one-shot mode finish_conn/2 closes
                             %% it.
+                            %% Log at ERROR level so log-level alerting still sees
+                            %% a permanent failure: before issue #80 short-circuited
+                            %% the retry loop, this failure would have exhausted the
+                            %% retries and reached the ?LOG_ERROR clause above.
+                            ?LOG_ERROR(
+                                "Request to AWS service has failed with a permanent error: ~tp",
+                                [Message]
+                            ),
                             State3 = finish_conn(Conn1, State1),
                             {error, exhausted_error(Error), State3};
                         retriable ->
--- a/src/aws_lib.erl
+++ b/src/aws_lib.erl
@@ -968,7 +968,11 @@ perform_request_reuse(Service, Method, Headers, Path, Body, Options, State0, Con
                     {aws_lib_response:format_response(Error), ConnSlot0, retriable}
             end;
         {error, _} = Error ->
-            {Error, ConnSlot0, retriable}
+            %% Shape the 2-tuple error through format_response/1 into the
+            %% 3-tuple result the retry loop matches on; returning the raw
+            %% {error, Reason} here would raise a case_clause in
+            %% api_request_with_retries/10.
+            {aws_lib_response:format_response(Error), ConnSlot0, retriable}
     end.
 
 %% Attach the connection slot and retry verdict to carry forward based on the
--- a/src/aws_lib.erl
+++ b/src/aws_lib.erl
@@ -28,7 +28,6 @@
     local_time/0,
     api_get_request/3,
     api_post_request/5,
-    status_text/1,
     open_connection/2, open_connection/3,
     close_connection/1,
     direct_request/7,
@@ -616,37 +615,6 @@ do_refresh_credentials(State0) ->
             {error, Reason}
     end.
 
--spec format_response(Response :: httpc_result()) -> result().
-%% @doc Format the httpc response result, returning the request result data
-%% structure. The response body will attempt to be decoded by invoking the
-%% maybe_decode_body/2 method.
-%% @end
-%% Any 2xx is a success. Every other status (3xx redirects we do not follow,
-%% 4xx, 5xx) is an error: gun is not configured to follow redirects, so a 3xx
-%% is a request we could not complete.
-format_response({ok, {{_Version, StatusCode, _Message}, Headers, Body}}) when
-    StatusCode >= 200, StatusCode < 300
-->
-    {ok, {Headers, maybe_decode_body(get_content_type(Headers), Body)}};
-format_response({ok, {{_Version, _StatusCode, Message}, Headers, Body}}) ->
-    {error, Message, {Headers, maybe_decode_body(get_content_type(Headers), Body)}};
-format_response({error, Reason}) ->
-    {error, Reason, undefined}.
-
--spec get_content_type(Headers :: headers()) -> {Type :: string(), Subtype :: string()}.
-%% @doc Fetch the content type from the headers and return it as a tuple of
-%%      {Type, Subtype}.
-%% @end
-get_content_type(Headers) ->
-    Value =
-        case proplists:get_value(<<"content-type">>, Headers, undefined) of
-            undefined ->
-                proplists:get_value(<<"Content-Type">>, Headers, "text/xml");
-            Other ->
-                Other
-        end,
-    parse_content_type(Value).
-
 -spec expired_credentials(Expiration :: calendar:datetime()) -> boolean().
 %% @doc Indicates whether the given expiration is at or within the refresh
 %% buffer window. Credentials are treated as expired a few minutes before they
@@ -671,47 +639,6 @@ expired_credentials(Expiration) ->
 %% @end
 local_time() ->
     calendar:universal_time().
-
--spec maybe_decode_body(ContentType :: {nonempty_string(), nonempty_string()}, Body :: body()) ->
-    list() | body().
-%% @doc Attempt to decode the response body by its MIME
-%% @end
-maybe_decode_body(_, <<>>) ->
-    <<>>;
-maybe_decode_body({"application", Subtype}, Body) ->
-    case is_json_subtype(Subtype) of
-        true -> aws_lib_json:decode(Body);
-        false -> maybe_decode_xml(Subtype, Body)
-    end;
-maybe_decode_body({_, Subtype}, Body) ->
-    maybe_decode_xml(Subtype, Body).
-
-maybe_decode_xml("xml", Body) ->
-    aws_lib_xml:parse(Body);
-maybe_decode_xml(_Subtype, Body) ->
-    Body.
-
-%% Recognise the JSON subtypes AWS services return. Covers plain `json', every
-%% AWS JSON protocol version (`x-amz-json-1.0', `x-amz-json-1.1', and any future
-%% `x-amz-json-*'), and the RFC 6839 structured `+json' suffix. Secrets Manager
-%% and other services use the JSON 1.1 protocol, so matching only 1.0/plain left
-%% those responses undecoded (issue #99).
-is_json_subtype("json") ->
-    true;
-is_json_subtype("x-amz-json-" ++ _Version) ->
-    true;
-is_json_subtype(Subtype) ->
-    lists:suffix("+json", Subtype).
-
--spec parse_content_type(ContentType :: string()) -> {Type :: string(), Subtype :: string()}.
-%% @doc parse a content type string returning a tuple of type/subtype
-%% @end
-parse_content_type(ContentType) when is_binary(ContentType) ->
-    parse_content_type(binary_to_list(ContentType));
-parse_content_type(ContentType) ->
-    Parts = string:tokens(ContentType, ";"),
-    [Type, Subtype] = string:tokens(lists:nth(1, Parts), "/"),
-    {Type, Subtype}.
 
 -spec expired_imdsv2_token('undefined' | imdsv2token()) -> boolean().
 %% @doc Determine whether or not an Imdsv2Token has expired.
@@ -887,7 +814,7 @@ api_request_with_retries(
 ) ->
     case ensure_credentials_valid(State0) of
         {ok, State1} ->
-            {Result, Conn1} =
+            {Result, Conn1, Verdict} =
                 perform_request_reuse(Service, Method, Headers, Path, Body, [], State1, Conn0),
             case Result of
                 {ok, {_Headers, Payload}, State2} ->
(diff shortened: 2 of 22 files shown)


lukebakken (commented):
Self-review notes, one inline comment per finding.

src/aws_lib.erl
```diff
@@ -1002,29 +943,47 @@ perform_request_reuse(Service, Method, Headers, Path, Body, Options, State0, Con
...
                 {error, _Reason} = Error ->
-                    {format_response(Error), ConnSlot0}
+                    %% A malformed URI never becomes valid on retry, but it is not
+                    %% an AWS response either; preserve the pre-existing retry
+                    %% behaviour by treating it as retriable.
+                    {aws_lib_response:format_response(Error), ConnSlot0, retriable}
             end;
         {error, _} = Error ->
-            {Error, ConnSlot0}
+            {Error, ConnSlot0, retriable}
```
lukebakken: This clause returns `prepare_signed_request`'s 2-tuple `{error, Reason}` as `Result`, but the `case Result of` in `api_request_with_retries` only matches `{ok, {_Headers, Payload}, State2}` and `{error, Message, Response}`, so the 2-tuple crashes with `case_clause`. Reachable via a malformed `AWS_ENDPOINT_URL` override: `sign_headers` returns `{error, {malformed_uri, URI}}` before the inner URI parse whose error IS wrapped by `format_response` in the clause above. Pre-existing on `main`, but this PR rewrote both the returning clause and the consuming `case`. No test covers this path.

src/aws_lib_response.erl
```diff
@@ -0,0 +1,242 @@
...
+-include("aws_lib.hrl").
+
+%% Substring markers, matched case-insensitively, that identify a throttling
+%% error code. This is botocore's _THROTTLED_ERROR_CODES
+%% (botocore/retries/standard.py) collapsed into lowercase substrings:
+%% `throttl' covers Throttling, ThrottlingException, ThrottledException,
+%% RequestThrottledException, RequestThrottled, and EC2ThrottledException;
+%% `limitexceeded' covers RequestLimitExceeded, BandwidthLimitExceeded, and
+%% LimitExceededException; the rest map one to one.
+-define(THROTTLING_MARKERS, [
```
lukebakken, line 33: This list covers botocore's `_THROTTLED_ERROR_CODES` but not its `_TRANSIENT_ERROR_CODES` (`RequestTimeout`, `RequestTimeoutException`; the third entry, `PriorRequestNotComplete`, is already here). Before this PR every error was retried, so an S3 HTTP 400 with `<Code>RequestTimeout</Code>` (a transient stalled-socket error that botocore retries) was retried and now fails on the first attempt. Adding `<<"requesttimeout">>` covers both codes as substrings.

src/aws_lib.erl
```diff
@@ -908,27 +835,40 @@ api_request_with_retries(
...
-                        Conn1,
-                        keep_informative_error(LastError, Error)
-                    )
+                    case Verdict of
+                        not_retriable ->
+                            %% A 4xx client error that will not succeed on retry
+                            %% (issue #80): close the connection and return the
+                            %% decoded error to the caller immediately rather than
+                            %% burning the remaining retries.
+                            close_if_open(Conn1),
```
lukebakken: In cross-request reuse mode this closes the shared connection but returns no updated state, so the caller's `reuse_conn` still holds the dead pid: the next request seeds it, `gun:await` surfaces `{error, {down, _}}` (classified retriable), and every post-4xx request burns one attempt plus a `timer:sleep` before reopening. The exhaustion path had the same pattern pre-PR, but this branch reaches it on the FIRST 4xx. After a cleanly completed 4xx exchange the connection is still usable; in reuse mode it should be handed back rather than closed.

test/aws_lib_tests.erl
```diff
@@ -739,6 +606,75 @@ api_get_request_test_() ->
...
+                    ])
+                ),
+
+                Result = aws_lib:api_get_request("AWS", "API", State),
+                ?assertMatch({ok, [{"data", "value"}], _State}, Result),
+                %% Two attempts: throttling was retried.
+                ?assertEqual(2, meck:num_calls(gun, await, '_')),
+                meck:validate(gun)
+            end},
             {"a decoded error survives a later transport failure", fun() ->
```
lukebakken, line 717: This test now passes vacuously: the 400 `InvalidParameterValue` first response is `not_retriable`, so the loop returns after one `gun:await` call and the four queued transport failures are never consumed. Nothing exercises `keep_informative_error/2`'s ordering clause any more (issue #82). Making the first response retriable (a 503 or a `ThrottlingException` 400) followed by the transport failures, and asserting the call count, restores the coverage.

src/aws_lib.erl
```diff
@@ -908,27 +835,40 @@ api_request_with_retries(
...
-                        Body,
-                        Headers,
-                        Retries - 1,
-                        WaitTime,
-                        State1,
-                        Conn1,
-                        keep_informative_error(LastError, Error)
-                    )
+                    case Verdict of
+                        not_retriable ->
```
lukebakken, line 842: Terminal permanent failures no longer emit any ERROR-level log. On `main` a 403 always exhausted retries and hit the `?LOG_ERROR("Request to AWS service has failed after ~b retries", ...)` clause; this early return logs only the two `?LOG_WARNING`s above. Worth an explicit `?LOG_ERROR` here so log-level alerting still sees permanent failures.

src/aws_lib_response.erl
```diff
@@ -0,0 +1,242 @@
...
+    retriable.
+
+%% Decide whether an error response indicates throttling. The error code is
+%% taken from the X-Amzn-Errortype header when present alongside the values
+%% stored under the AWS error-code body keys; matching the error code rather
+%% than the whole body keeps a marker inside a message or resource name from
+%% triggering a retry. A body that could not be decoded (still a raw binary)
+%% falls back to a substring search of the raw bytes.
+-spec is_throttling_error(headers(), list() | body()) -> boolean().
+is_throttling_error(Headers, Body) when is_binary(Body) ->
```
lukebakken, line 125: The raw-body fallback substring-searches the whole body, reintroducing for `text/plain` and `text/html` responses the false-positive class the decoded path deliberately excludes: a proxy or ALB 4xx page whose text contains `throttl` retries the full budget on a permanent error. The cost is bounded (latency only, the correct error is still returned) and the tests encode the tradeoff, but consider narrowing the fallback to the `X-Amzn-Errortype` header plus the status rules, or documenting the tradeoff in the comment above.

src/aws_lib_httpc.erl
```diff
@@ -109,15 +109,17 @@ request(Conn, Method, Path, Headers, Body, Timeout) ->
         StreamRef = do_gun_request(Conn, Method, Path, HeadersBin, Body),
         case gun:await(Conn, StreamRef, Timeout) of
             {response, fin, Status, RespHeaders} ->
-                {ok, {{http_version, Status, aws_lib:status_text(Status)}, RespHeaders, <<>>}};
+                {ok, {
+                    {http_version, Status, aws_lib_response:status_text(Status)}, RespHeaders, <<>>
```
lukebakken: This call makes the dependency mutual: `aws_lib_httpc` calls `aws_lib_response:status_text/1` while `aws_lib_response` calls `response_status/1` and types against `response()`, contradicting the one-way "httpc produces, response interprets" contract in the new module's doc. `status_text/1` is response construction (its own comment: to build the status line Gun does not carry), so it belongs here in `aws_lib_httpc`, unexported.

include/aws_lib.hrl
```diff
@@ -163,10 +163,12 @@
     | {ok, {status_code(), body()}}
     | {error, term()}.
```
lukebakken: `httpc_result()` is dead after this PR: its only live consumer was `format_response/1`'s old spec, which now uses `aws_lib_httpc:response()`, and the only remaining reference is a commented-out spec in `aws_lib_config.erl`. Delete the type and that stale commented spec so `aws_lib_httpc:response()` is the single response type.

src/aws_lib_response.erl
```diff
@@ -0,0 +1,242 @@
...
+            %% nothing.
+            error_code_values(Value)
+    end.
+
+%% Case-insensitive substring match of the throttling markers against an error
+%% code or a raw body. to_utf8_binary/1 keeps multi-byte codepoints intact
+%% where a bare iolist_to_binary/1 would fail on non-latin1 input; a value that
+%% is not chardata at all (e.g. a nested proplist under an error-code key)
+%% makes it throw, which simply means "not throttling".
+contains_throttling_marker(Value) ->
```
lukebakken, line 168: This duplicates the `try string:lowercase(rabbit_data_coercion:to_utf8_binary(X))` pattern from `is_errortype_header/1`, and the `_ -> false` of-clause is unreachable: `string:lowercase/1` on a binary always returns a binary, and a non-binary coercion result raises into the catch. A single shared safe-lowercase helper with a plain catch would serve both call sites and drop the dead clause.

src/aws_lib_response.erl
```diff
@@ -0,0 +1,242 @@
...
+            undefined ->
+                proplists:get_value(<<"Content-Type">>, Headers, "text/xml");
+            Other ->
+                Other
+        end,
+    parse_content_type(Value).
+
+-spec maybe_decode_body(ContentType :: {nonempty_string(), nonempty_string()}, Body :: body()) ->
+    list() | body().
+%% @doc Attempt to decode the response body by its MIME
```
lukebakken: Grammar per the repo comment rules: this `@doc` trails off without a full stop ("by its MIME" what?), and the `@doc` on `parse_content_type/1` below lacks a capital letter, articles, and a full stop. Both were moved verbatim, but the move into a new file is the moment to fix them.

src/aws_lib.erl
lukebakken: Resolved: this clause now shapes the 2-tuple through `aws_lib_response:format_response/1` into the `{error, Reason, undefined}` 3-tuple the retry loop matches, matching the sibling URI-parse-error path directly above. Added `signing_failure_retries_test_/0`, which mecks `aws_lib_sign:headers/2` to fail and asserts the request exhausts retries cleanly; it fails with a `case_clause` when the fix is reverted.

src/aws_lib_response.erl
lukebakken: Resolved: added `<<"requesttimeout">>` to `?THROTTLING_MARKERS`, which covers both transient codes `RequestTimeout` and `RequestTimeoutException` as substrings. The doc comment now notes the list also covers botocore's `_TRANSIENT_ERROR_CODES`. Extended the "botocore throttling code families are covered" test with `RequestTimeout`, `RequestTimeoutException`, and `PriorRequestNotComplete`; verified it fails when the marker is removed.

src/aws_lib.erl
lukebakken: Resolved: the retry loop now threads the aws_state() out on every error return, and a new `finish_conn/2` (generalized from the old `finish_conn_success/2`) reconciles the connection into that state at every terminal path. After a cleanly completed not_retriable 4xx the connection is handed back via the state's `reuse_conn` in reuse mode (kept as usable as after a 2xx) and closed in one-shot mode; a transport-dropped slot records `none` rather than `undefined` so the state stays in reuse mode. The error return type changed from `{error, term()}` to `{error, term(), aws_state()}`, cascading through `aws_s3`/`aws_sms`/`aws_acm_pca`/`aws_arn_util`/`aws_arn_config`/`aws_arn_config_oauth2` and the `aws_auth_validate_ssl` adapter; `aws_iam` and `instance_volumes/1` collapse it back to their existing 2-tuple contracts. Added a reuse-mode regression test asserting the connection is not closed and lands in `reuse_conn`; verified it fails against the old close-on-4xx behavior. Full suite (557 eunit) green, xref clean, dialyzer unchanged from baseline.

src/aws_lib.erl
lukebakken: Resolved in d2996c7. Added an explicit `?LOG_ERROR` on the `not_retriable` short-circuit path so a permanent failure (e.g. a 403) is still visible to log-level alerting, as it was before issue #80 short-circuited the retry loop. A comment records why. No test was added: the suite has no log-capture harness and this is a logging-only change on a path already exercised by the reuse-mode and not_retriable tests.

test/aws_lib_tests.erl
lukebakken: Resolved in 54ae61a. Changed the first response from a not_retriable 400 to a retriable 503 (with a decoded error body) so the loop keeps retrying and consumes the four queued transport failures, which restores coverage of `keep_informative_error/2`'s ordering clause. Also added `?assertEqual(?MAX_RETRIES, meck:num_calls(gun, await, '_'))` so the test asserts all attempts are consumed rather than passing after one. Verified non-vacuity: breaking the ordering clause now fails the test; restored, all 50 pass.

src/aws_lib_response.erl
lukebakken: Resolved in ad3b7a9 by narrowing the fallback (your first suggestion). An undecodable body (text/plain, text/html) is no longer substring-searched: it carries no structured error code, so classification rests on the `X-Amzn-Errortype` header plus the 5xx/429 status rules. This removes the false-positive class entirely, so a proxy or ALB 4xx error page containing `throttl` is no longer retried on a permanent error. Rewrote the "undecodable body" test to assert the raw text is not searched, and added a companion test showing such a body still retries when the throttling code arrives in the header. Verified: reverting the narrowing fails the new test; restored, all 33 response tests pass, xref clean, dialyzer unchanged from baseline.

src/aws_lib_httpc.erl
lukebakken: Resolved in e1267c8. Moved `status_text/1` out of `aws_lib_response` (removed the export and definition) and into `aws_lib_httpc` as an unexported internal function, updating the two call sites to call it directly. This restores the one-way "httpc produces, response interprets" contract: `aws_lib_httpc` no longer calls into `aws_lib_response`. As you noted, `status_text/1` is response construction (it fills in the reason phrase Gun does not carry), so it belongs alongside the rest of the `response/0` shaping.

I also checked whether the hand-rolled table could be dropped in favour of OTP's `httpd_util:reason_phrase/1` (inets is already a LOCAL_DEP). Decided against it: `httpd_util` maps every unknown code to "Internal Server Error", so a 429 (a code the classifier treats as retriable throttling) would be mislabelled in logs, whereas the current fallback surfaces the numeric code. Its known phrases also differ (404 -> "Object Not Found", 416 -> "Requested Range Not Satisfiable"), which would churn tests for no functional gain since nothing matches on the phrase text.

All 96 tests pass (httpc, response, lib suites), xref clean, dialyzer unchanged from baseline.

include/aws_lib.hrl
lukebakken: Resolved in d369d4c. Deleted the `httpc_result()` type and the stale commented-out `-spec parse_az_response(httpc_result())` in `aws_lib_config`, so `aws_lib_httpc:response()` is now the single response type. Kept the `@doc` on `parse_az_response/1` in the style of the sibling `parse_body_response/1`. Verified the full build is clean (no unused-type warning), all 70 config tests pass, xref clean, dialyzer unchanged from baseline. Left `status_line()` in place: after removing `httpc_result()` it is referenced only in `aws_lib_httpc`'s doc comments as the type the `response()` shape deliberately differs from.

src/aws_lib_response.erl
lukebakken: Resolved in 702cac9. Extracted a shared `safe_lowercase/1` helper (the `try string:lowercase(to_utf8_binary(X))` pattern with a plain catch, returning `{ok, Binary}` or `error`) and routed both `is_errortype_header/1` and `contains_throttling_marker/1` through it. Dropped the unreachable `_ -> false` of-clause: as you noted, `string:lowercase/1` on a binary always returns a binary, and a non-chardata coercion raises into the catch. The coercion rationale now lives once, on the helper. All 33 response tests pass, xref clean, dialyzer unchanged from baseline.

src/aws_lib_response.erl
lukebakken: Resolved in 715451c. Fixed both: `maybe_decode_body/2`'s `@doc` now reads "Attempt to decode the response body according to its MIME content type." (completes the trailing sentence), and `parse_content_type/1`'s now reads "Parse a content type string, returning a tuple of the type and subtype." (capital, articles, full stop). The `get_content_type/1` `@doc` just above was already a complete sentence with a full stop, so it was left as is.

the-mikedavis (approved):

PR #141, opened 2026-07-13
Guard HTTP validation divergence against real auth backend

sdewhitt:
## Summary

The HTTP validation method (`aws_auth_validate_http`) mirrors `rabbit_auth_backend_http` by hand: it builds the same request queries and interprets the same `allow`/`deny` responses, so a green validation result reflects how the real broker would behave. Nothing pinned that contract, so an upstream change to the query encoding or the response grammar could make the probe diverge from a real broker request without any test noticing.

This adds that protection, analogous to the LDAP query-parser parity tests (`aws_auth_validate_tests:parity_test_/0`), which exist for the same reason: to keep a hand-mirrored contract from silently drifting.

## What changed

- Add `rabbitmq_auth_backend_http` as a `TEST_DEP`.
- New `test/aws_auth_validate_http_parity_tests.erl`:
  - **Query-encoding parity** -- the probe's per-path query strings encode identically to the exported `rabbit_auth_backend_http:q/1`. This is a true differential test: both are run on the same params and asserted equal, so a change in upstream encoding (e.g. space as `%20` instead of `+`) fails the build.
  - **`join_tags/1` parity** -- against the exported upstream helper.
  - **Response-grammar contract** -- `classify_response/2` accepts exactly the `allow`/`deny` forms the broker's parsing accepts (bare `deny` or `allow` + space-separated tags for authn; bare `allow`/`deny` for authz), and rejects the rest. The broker's parsing is inline in `user_login_authentication/2` and `req/2` (not an exported pure function), so this side is pinned against the documented grammar rather than called directly.
- Expose `query_for/1` and `params_for/1` under `-ifdef(TEST)` for the parity test.
- Fix a stale comment: the grammar comment referenced a `parse_resp/1` that does not exist upstream; the parsing lives in `user_login_authentication/2` + `req/2`. (This stale reference is itself an example of the drift this PR guards against.)

The differential checks skip -- rather than fail -- when upstream is not usable in the node (module or its deps not loaded), matching how the LDAP parity tests handle the RMQ-version CI matrix.

## Limitation

This guard is necessarily weaker than the LDAP parity tests. LDAP reimplements a whole parser and gets a clean "run both implementations, assert equal" test. HTTP only gets query-encoding parity plus grammar-contract assertions, because the two things most worth pinning -- upstream's `?SUCCESSFUL_RESPONSE_CODES` macro and its inline `allow`/`deny` parser -- are not exported. A full parity test would require a small upstream change to export a response-code predicate and factor out the response parser.

## Testing

- Full eunit suite passes (603 tests).
- Confirmed the differential checks are non-vacuous: the upstream-usable guards return true in the umbrella build, so the query-encoding and `join_tags` parity tests run against the real backend rather than skipping.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ define PROJECT_ENV
 endef
 
 DEPS = rabbit_common rabbit rabbitmq_management gun
-TEST_DEPS = meck proper rabbitmq_ct_helpers rabbitmq_ct_client_helpers rabbitmq_auth_backend_ldap
+TEST_DEPS = meck proper rabbitmq_ct_helpers rabbitmq_ct_client_helpers rabbitmq_auth_backend_ldap rabbitmq_auth_backend_http
 LOCAL_DEPS = crypto inets ssl xmerl public_key eldap
 
 PLT_APPS = rabbit
--- a/src/aws_auth_validate_http.erl
+++ b/src/aws_auth_validate_http.erl
@@ -689,7 +689,14 @@ with_default_sni(SslOpts, Host) ->
 %% fields rabbit_auth_backend_http sends so a conformant server returns a 2xx
 %% (allow or deny) rather than a 400 for missing params. The username is an
 %% obvious placeholder so an operator reading their auth-server logs can see it
-%% was a validation probe, not a real login attempt.
+%% was a validation probe, not a real login attempt. The authz paths carry a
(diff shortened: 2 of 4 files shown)


the-mikedavis (approved):
👍

lukebakken (approved):

PR #146, opened 2026-07-15
Add optional OAuth authorization-evaluation layer to validation endpoint

sdewhitt:
Follow-up enhancement for *Issue #144*

## Summary

Base OAuth validation proves reachability + JWKS well-formedness and (when a customer supplies their own `access_token`) signature + `exp`/`aud`. It stops at **authentication**: it never answers "would the broker grant this principal a permission?" That authorization gap is the #1 source of "my token is broken (it isn't)" support tickets, where the real fault is a `scope_prefix` / `scope_aliases` typo, not the token.

This adds an **optional** authorization-evaluation step that closes the gap without a broker restart and without accepting any secret. It is a **triage aid**, not an authorization gate: it localizes *which* layer of an OAuth config is wrong (reachability vs. authentication vs. authorization) and, for authz, names the failing stage.

## How it works

`aws_auth_validate_oauth_authz` (new): given an already-verified token's claims and a customer-supplied `authz_check` `{vhost, resource, permission}`, it answers whether the broker would grant it. Implemented as a **runtime soft dependency** on `rabbitmq_auth_backend_oauth2`:

- It **calls the backend's own pure decision functions** (`normalize_token_scope/2`, `get_expanded_scopes/3`, `rabbit_oauth2_scope:resource_access/4`) when-and-only-when the module is loaded, guarded by `code:ensure_loaded/1` + `function_exported/3`.
- It is **not** added to `DEPS`, so aws-plugin users who never run OAuth are unaffected.
- Because it **executes the broker's code rather than mirroring it, drift is impossible by construction** -- there is nothing to keep in sync.
- When the backend is absent, authz is unavailable and says so (`config_conflict`); it never falls back to a home-grown matcher.

## Example request

```
PUT /api/aws/auth/validate/oauth
{
  "jwks_uri": "https://idp.example.com/.well-known/jwks.json",
  "access_token": "<customer-minted JWT>",
  "resource_server_id": "rabbitmq",
  "scope_prefix": "rabbitmq.",
  "authz_check": { "vhost": "/", "resource": "my-queue", "permission": "read" }
}
```

`204` = the broker would grant it. `422 authz_unverified` = it would not (with a message naming the failing stage). The `access_token` / `authz_check` fields are entirely optional; omitting them leaves the base reachability behaviour unchanged.

## Non-goals / out of scope (unchanged by this PR)

- This does **not** turn the endpoint into an authorization gate; the base check still asserts reachability + authentication only.
- `assume_role` from request input remains **excluded** (confused-deputy); no change here.
- No broker-side `client_credentials` grant -- the customer-supplied token path already gives end-to-end confidence with no secret transiting the endpoint.

## Changes

- **`aws_auth_validate_oauth`:** wire the optional layer in. New `allowed_fields` (`scope_aliases`, `additional_scopes_key`, `scope_prefix`, `scope_pattern_syntax`, `authz_check`); pure-phase shape checks that require an `access_token` and reject a malformed check before any network/crypto; `verify_token_claims/3` now returns the decoded claims so `maybe_check/2` can run against them after verification.
- **`Makefile`:** add `rabbitmq_auth_backend_oauth2` to `TEST_DEPS` so the layer's decision functions are loadable under test.

## Testing

- **CT (`aws_auth_validate_mgmt_SUITE`): 22/22** against the umbrella (`gmake`), including two new cases that drive the authz layer **end-to-end through the Cowboy handler** -- dispatch is **not** mocked; only the JWKS network seam is stubbed, and the token + JWKS are minted on the broker node. `oauth_authz_grants_returns_204` and `oauth_authz_denies_returns_422` use the **same** verified token (grants `read`) and differ only in the requested permission, so the 204-vs-422 split proves the `authz_check` field survives the registry allowlist filter and the config is genuinely consumed -- not a no-op.
- **eunit:** availability probe; direct-scope grant; regex-syntax grant; present-scope-no-match and no-effective-scopes-after-prefix denials (asserting the distinct messages); IAM `scope_aliases` expansion grant; `authz_check`-without-token and bad-permission rejections.
- Both the eunit and CT authz cases **skip gracefully** when `rabbitmq_auth_backend_oauth2` is not loaded on the node (mirroring the LDAP suite's slapd skip). With it in `TEST_DEPS` the backend is loaded in CI, so the cases run rather than skip.
- `erlfmt -c` clean; clean release build.

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,16 +20,18 @@ DEP_PLUGINS = rabbit_common/mk/rabbitmq-plugin.mk
 include ../../rabbitmq-components.mk
 include ../../erlang.mk
 
-# The optional OAuth authz-evaluation layer (aws_auth_validate_oauth_authz)
-# builds the broker's #resource_server{} record and so needs oauth2.hrl at
-# compile time. rabbitmq_auth_backend_oauth2 is a RUNTIME soft dependency, and
-# its header is absent on older broker series (e.g. 3.13.x). Define
-# HAVE_OAUTH2_RESOURCE_SERVER only when the header actually exists in the
-# resolved deps dir, so the plugin still compiles where it does not (the authz
-# layer is then merely unavailable at runtime via available/0). This must come
-# after erlang.mk so DEPS_DIR and ERLC_OPTS are set.
+# Gate the optional OAuth authz layer on the arity-4 scope API it needs, not
+# just oauth2.hrl existing: the header predates resource_access/4 and the
+# scope_pattern_syntax field (both landed in the v4.2.0-beta series -- the
+# supported-series floor), so a header-only guard would build against a missing
+# function. scope_pattern_syntax in the resolved header is the sentinel for that
+# API. When absent, the module still compiles (-else branch), available/0
+# returns false, and authz_check reports config_conflict. Must follow erlang.mk
+# so DEPS_DIR and ERLC_OPTS are set.
 OAUTH2_HRL = $(DEPS_DIR)/rabbitmq_auth_backend_oauth2/include/oauth2.hrl
 ifneq ($(wildcard $(OAUTH2_HRL)),)
+ifneq ($(shell grep -l scope_pattern_syntax $(OAUTH2_HRL) 2>/dev/null),)
 ERLC_OPTS += -DHAVE_OAUTH2_RESOURCE_SERVER=1
 TEST_ERLC_OPTS += -DHAVE_OAUTH2_RESOURCE_SERVER=1
+endif
 endif
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ define PROJECT_ENV
 endef
 
 DEPS = rabbit_common rabbit rabbitmq_management gun jose
-TEST_DEPS = meck proper rabbitmq_ct_helpers rabbitmq_ct_client_helpers rabbitmq_auth_backend_ldap rabbitmq_auth_backend_http
+TEST_DEPS = meck proper rabbitmq_ct_helpers rabbitmq_ct_client_helpers rabbitmq_auth_backend_ldap rabbitmq_auth_backend_http rabbitmq_auth_backend_oauth2
 LOCAL_DEPS = crypto inets ssl xmerl public_key eldap
 
 PLT_APPS = rabbit
--- a/Makefile
+++ b/Makefile
@@ -19,3 +19,17 @@ DEP_PLUGINS = rabbit_common/mk/rabbitmq-plugin.mk
 
 include ../../rabbitmq-components.mk
 include ../../erlang.mk
+
+# The optional OAuth authz-evaluation layer (aws_auth_validate_oauth_authz)
+# builds the broker's #resource_server{} record and so needs oauth2.hrl at
+# compile time. rabbitmq_auth_backend_oauth2 is a RUNTIME soft dependency, and
+# its header is absent on older broker series (e.g. 3.13.x). Define
+# HAVE_OAUTH2_RESOURCE_SERVER only when the header actually exists in the
+# resolved deps dir, so the plugin still compiles where it does not (the authz
+# layer is then merely unavailable at runtime via available/0). This must come
+# after erlang.mk so DEPS_DIR and ERLC_OPTS are set.
+OAUTH2_HRL = $(DEPS_DIR)/rabbitmq_auth_backend_oauth2/include/oauth2.hrl
+ifneq ($(wildcard $(OAUTH2_HRL)),)
+ERLC_OPTS += -DHAVE_OAUTH2_RESOURCE_SERVER=1
+TEST_ERLC_OPTS += -DHAVE_OAUTH2_RESOURCE_SERVER=1
+endif
--- a/src/aws_auth_validate_oauth_authz.erl
+++ b/src/aws_auth_validate_oauth_authz.erl
@@ -298,14 +298,15 @@ decide(Scopes, Resource, PermAtom, Syntax) ->
         false -> {error, authz_unverified, ?REASON_AUTHZ_NO_MATCH}
     end.
 
-%% Map the supplied permission string to the broker's permission atom. Only the
-%% three RabbitMQ permission verbs are accepted; anything else is bad input
-%% (caught by the caller's is_binary guard returning a non-matching atom is not
-%% possible, so we constrain here).
+%% Map the supplied permission string to the broker's permission atom. The
+%% permission is already allowlisted to exactly these three verbs in the pure
+%% phase (aws_auth_validate_oauth:parse_authz_check/3), so this clause set is
+%% total over every value that can reach it. It has no catch-all: an unexpected
+%% value must fail loudly here rather than be silently mapped to a non-matching
+%% atom that resource_access/4 would report as a spurious "none grant".
 to_permission_atom(<<"configure">>) -> configure;
 to_permission_atom(<<"write">>) -> write;
-to_permission_atom(<<"read">>) -> read;
-to_permission_atom(_) -> undefined.
+to_permission_atom(<<"read">>) -> read.
 
 -else.
 
--- a/src/aws_auth_validate_oauth_authz.erl
+++ b/src/aws_auth_validate_oauth_authz.erl
@@ -61,13 +61,28 @@
 
 -export([maybe_check/2, available/0]).
 
-%% We build the broker's #resource_server{} record, so we include its header.
-%% This is a COMPILE-TIME header include (a record shape), not a code/DEPS
-%% dependency: if upstream changes the record, this fails loudly at build time
-%% rather than drifting silently at runtime. #resource{} is from rabbit_common,
-%% already a real dependency of this plugin.
--include_lib("rabbitmq_auth_backend_oauth2/include/oauth2.hrl").
+%% The evaluation path builds the broker's #resource_server{} record, so it needs
+%% that record's shape (from oauth2.hrl). But rabbitmq_auth_backend_oauth2 is a
+%% RUNTIME soft dependency, not a build DEPS: the plugin is compiled against a
+%% range of broker series, and the oauth2 backend's shape is NOT stable across
+%% them -- older series (e.g. 3.13.x) ship no oauth2.hrl at all, and newer ones
+%% add record fields (scope_pattern_syntax). Hard-including the header would make
+%% the whole plugin fail to COMPILE on any series that lacks it, defeating the
+%% soft-dependency design (the feature is meant to be merely UNAVAILABLE there,
+%% via available/0, not a build break).
+%%
+%% So the header include and the record-using evaluation code are guarded by
+%% -ifdef(HAVE_OAUTH2_RESOURCE_SERVER), a macro the Makefile defines only when
+%% $(DEPS_DIR)/rabbitmq_auth_backend_oauth2/include/oauth2.hrl actually exists at
+%% build time. When it is absent, the module still compiles; available/0 returns
+%% false and maybe_check/2 reports config_conflict (authz unavailable). When it
+%% is present, the record-typed evaluator is compiled in. We do NOT reference
+%% fields that vary across supported series (scope_pattern_syntax): the syntax is
+%% threaded as a plain argument to the scope functions instead (see evaluate/3).
 -include_lib("rabbit_common/include/resource.hrl").
+-ifdef(HAVE_OAUTH2_RESOURCE_SERVER).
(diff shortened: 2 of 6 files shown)


the-mikedavis (approved):

src/aws_auth_validate_oauth_authz.erl
```diff
@@ -0,0 +1,319 @@
...
+%% we actually want -- and stays false only when the backend genuinely is not
+%% deployed on this broker.
+-spec available() -> boolean().
+-ifdef(HAVE_OAUTH2_RESOURCE_SERVER).
+available() ->
+    module_ready(?SCOPE_MOD) andalso
+        module_ready(?OAUTH2_MOD) andalso
+        module_ready(?RS_MOD) andalso
+        erlang:function_exported(?SCOPE_MOD, resource_access, 4) andalso
+        erlang:function_exported(?OAUTH2_MOD, get_expanded_scopes, 3) andalso
```
lukebakken, line 189: **Blocking (portability).** These probes require `resource_access/4` and `get_expanded_scopes/3`, which landed later than `oauth2.hrl`. On a broker series that ships the header but only the arity-3/arity-2 API, this `andalso` chain short-circuits to false, so `available/0` is false even though the oauth2 backend is loaded -- and every `authz_check` returns the "plugin not loaded" `config_conflict`, which is false. `gmake xref` against such a tree fails outright: `Undefined function rabbit_oauth2_scope:resource_access/4`. Gate on the API (or the `scope_pattern_syntax` record field) rather than header presence, and pin the supported-series floor.

Makefile
```diff
@@ -19,3 +19,17 @@ DEP_PLUGINS = rabbit_common/mk/rabbitmq-plugin.mk
...
+# builds the broker's #resource_server{} record and so needs oauth2.hrl at
+# compile time. rabbitmq_auth_backend_oauth2 is a RUNTIME soft dependency, and
+# its header is absent on older broker series (e.g. 3.13.x). Define
+# HAVE_OAUTH2_RESOURCE_SERVER only when the header actually exists in the
+# resolved deps dir, so the plugin still compiles where it does not (the authz
+# layer is then merely unavailable at runtime via available/0). This must come
+# after erlang.mk so DEPS_DIR and ERLC_OPTS are set.
+OAUTH2_HRL = $(DEPS_DIR)/rabbitmq_auth_backend_oauth2/include/oauth2.hrl
+ifneq ($(wildcard $(OAUTH2_HRL)),)
+ERLC_OPTS += -DHAVE_OAUTH2_RESOURCE_SERVER=1
```
lukebakken, line 34: **Blocking (root cause of the portability gap).** `HAVE_OAUTH2_RESOURCE_SERVER` is defined purely on `oauth2.hrl` existing, but the code compiled in behind it needs the arity-4 scope API and the `scope_pattern_syntax` field, which are newer than the header. Header-present does not imply API-present, so the guard admits a series where the evaluator compiles but every call fails `available/0`. Condition the macro on the arity-4 functions (or the record field), not just the header.

test/aws_auth_validate_oauth_tests.erl
```diff
@@ -832,6 +836,251 @@ access_token_no_leak_test_() ->
...
+            <<"permission">> => <<"read">>
+        }
+    },
+    ?assertMatch({error, input_invalid, _}, aws_auth_validate_oauth:validate(Body)).
+
+%% Run Fun only if the oauth2 backend is loaded; otherwise return an empty test
+%% list (graceful skip), mirroring how the LDAP integration suite skips without
+%% slapd.
+maybe_skip_authz(Fun) ->
+    case aws_auth_validate_oauth_authz:available() of
```
lukebakken, line 1141: **Blocking (coverage gap that hides #1).** When `available/0` is false the whole authz suite skips to `[]` and CI stays green with zero coverage -- which is exactly what masked the arity mismatch. Distinguish "backend genuinely absent" (a legitimate skip) from "backend present but API absent" (should be a hard failure), so a portability regression cannot pass silently.

test/aws_auth_validate_mgmt_SUITE.erl
```diff
@@ -217,6 +234,40 @@ init_per_testcase(oauth_response_no_secret = TC, Config) ->
...
+            {Token, JwksJson} = rabbit_ct_broker_helpers:rpc(
+                Config, 0, aws_auth_validate_oauth_test_helpers, oauth_authz_mint, []
+            ),
+            ok = rabbit_ct_broker_helpers:rpc(
+                Config, 0, ?MODULE, mock_oauth_network, [JwksJson]
+            ),
+            Config1 = rabbit_ct_helpers:set_config(Config, [{oauth_token, Token}]),
+            rabbit_ct_helpers:testcase_started(Config1, TC);
+        false ->
+            {skip, "rabbitmq_auth_backend_oauth2 is not loaded on the broker node"}
```
lukebakken: **Blocking (same coverage gap, CT side).** This `{skip, ...}` fires whenever `oauth2_backend_available/0` is false, including the present-backend-but-absent-API case. A skip there reads as "green" while the feature is entirely unexercised. Make the present-but-unusable case fail rather than skip.

src/aws_auth_validate_oauth_authz.erl
```diff
@@ -0,0 +1,319 @@
...
+
+maybe_check_available(Params, Check, Claims) ->
+    case available() of
+        false ->
+            {error, config_conflict, ?REASON_AUTHZ_UNAVAILABLE};
+        true ->
+            evaluate(Params, Check, Claims)
+    end.
+
+evaluate(Params, Check, Claims) ->
```
lukebakken, line 288: **Non-blocking (altitude).** `evaluate/3` replays the broker's decision as three separate internal calls (`new_resource_server/1` -> `normalize_token_scope/2` -> `get_expanded_scopes` -> `resource_access`). The matcher itself can't drift, but this sequencing can: any stage the backend adds to normalization must be re-mirrored here or the endpoint under-reports. Worth a comment flagging the coupling, or checking whether a single higher-level entry point fits.

src/aws_auth_validate_oauth_authz.erl
```diff
@@ -0,0 +1,319 @@
...
+%% the latter returns false for a module whose beam is on the path but has not
+%% yet been loaded into the VM, which would make availability depend on load
+%% order (a request could see the oauth2 backend as "absent" simply because
+%% nothing had called it yet this boot). ensure_loaded triggers the load if the
+%% beam is present, so this reflects "is the backend installed" -- the property
+%% we actually want -- and stays false only when the backend genuinely is not
+%% deployed on this broker.
+-spec available() -> boolean().
+-ifdef(HAVE_OAUTH2_RESOURCE_SERVER).
+available() ->
```
lukebakken, line 184: **Non-blocking (efficiency).** `available/0` runs `code:ensure_loaded` x3 plus several `function_exported/3` on every request. Low impact given the semaphore-bounded endpoint, but easily memoized -- cache only a `true` result (never `false`), so a backend that loads after the first call is still picked up on a later request.

lukebakken (changes requested):
This adds a genuinely useful authorization-evaluation layer, and the core design (executing the broker's own scope-decision functions rather than mirroring them) is the right instinct. I verified the happy path end-to-end against a freshly built main: `gmake xref` passes, `available/0` returns true, and the eunit plus CT authz cases exercise the real broker scope functions and pass (94 eunit, the two CT authz cases green).

One blocking issue on cross-series portability, one blocking test-coverage gap that hides it, and two non-blocking notes.

## Blocking

**1. The build guard keys on the header, not the API it actually needs.** `HAVE_OAUTH2_RESOURCE_SERVER` is defined whenever `oauth2.hrl` exists, but the evaluator requires the arity-4 scope API (`rabbit_oauth2_scope:resource_access/4`, `rabbit_auth_backend_oauth2:get_expanded_scopes/3`) and the `scope_pattern_syntax` record field, which landed later than that header. On a broker series where the header exists but the arity-4 API does not, the evaluator still compiles in, `available/0` returns false, and every `authz_check` is answered "requires the rabbitmq_auth_backend_oauth2 plugin to be loaded; it is not" -- a flatly false message, since the plugin is loaded. Confirmed by running `gmake xref` against such a tree: it fails with `Undefined function rabbit_oauth2_scope:resource_access/4`. Please guard on the API (`function_exported`) or the record field, and set the supported-series floor explicitly.

**2. The skip that hides issue 1.** When the oauth2 backend is present but the API is absent, both the eunit `maybe_skip_authz` and the CT `init_per_testcase` skip silently, so the whole grant/deny/alias/regex suite evaporates and CI stays green with zero coverage of the feature. A present-backend-but-absent-API state should be a hard failure, not a silent skip.

## Non-blocking

**3. `evaluate/3` hand-chains three broker internals.** The matcher can't drift (it's the broker's own code), but this call sequence can -- any normalization stage the backend adds must be re-mirrored here. Worth a note on the coupling, or exploring a higher-level entry point.

**4. `available/0` recomputes on every request.** Low impact on a semaphore-bounded endpoint, but memoizable -- cache only a `true` result so a backend that loads after first call is still picked up.

## Already addressed on this branch

- `authz_check.vhost` was accepted without validation; a non-binary value reached the broker matcher and crashed, misreported as a connection failure. Now rejected in the pure phase -- 6e890a1
- Removed a dead `to_permission_atom` catch-all that would have silently produced a non-matching permission instead of failing loudly -- 38694e3
- Extracted duplicated JWT-minting and request-URL test helpers into one shared module -- 20a0bd5

src/aws_auth_validate_oauth_authz.erl
sdewhitt: Done in 6fa88f7. available/0 already probed the arity-4 functions, but the real problem was upstream of it -- the module was being compiled in on series that lack the API. Fixed at the build guard (see the Makefile thread); this probe is unchanged but now only runs when the evaluator is actually compiled in.

Makefile
sdewhitt: Fixed in 6fa88f7. The macro now requires scope_pattern_syntax to be present in the resolved oauth2.hrl, not just the header to exist:

```
ifneq ($(wildcard $(OAUTH2_HRL)),)
ifneq ($(shell grep -l scope_pattern_syntax $(OAUTH2_HRL) 2>/dev/null),)
ERLC_OPTS += -DHAVE_OAUTH2_RESOURCE_SERVER=1
```

I confirmed via git log -S that scope_pattern_syntax and resource_access/4 landed in the same commit ("Support regexp or wildcard in scopes"), so the field is a reliable sentinel for the arity-4 API. That commit is on the v4.2.0-beta series -- documented as the supported-series floor in the Makefile comment. On an older series the module now compiles via the -else branch, available/0 is false, and authz_check reports config_conflict. Verified gmake xref passes on both a floor+ and a pre-floor tree.

test/aws_auth_validate_oauth_tests.erl
sdewhitt: Added evaluator_compiled_in/0 -- the compile-time counterpart to the runtime available/0. 

The hard failure now fires only when the evaluator was compiled in AND the backend is loaded, yet available/0 is still false -- i.e. the genuine portability regression. A pre-floor build skips legitimately again. I verified by compiling the module without the macro: evaluator_compiled_in=false, so it skips rather than fails.

test/aws_auth_validate_mgmt_SUITE.erl
sdewhitt: Same fix applied on the CT side in 6fa88f7/53a89f6. init_per_testcase now RPCs evaluator_compiled_in/0 + backend_loaded/0 to the broker node: ct:fail only when the evaluator was compiled in and the backend is loaded but available/0 is false; otherwise {skip, ...}. On main both authz cases run and pass (204/422); on pre-floor series they skip cleanly.

src/aws_auth_validate_oauth_authz.erl
sdewhitt: Added a coupling note above evaluate/3 in 6fa88f7 flagging that the new_resource_server -> normalize_token_scope -> get_expanded_scopes -> resource_access sequencing can drift if the backend inserts a normalization stage, and pointing at a single higher-level entry point as the preferred fix if one is ever exposed. Kept it as a comment rather than restructuring, since there's no such entry point today.

src/aws_auth_validate_oauth_authz.erl
sdewhitt: Done in 6fa88f7. available/0 now memoizes in persistent_term, caching only a true result -- never false -- so a backend enabled after the first call is still picked up on a later request. The probe runs at most once per boot on the request path.

PR #147, opened 2026-07-15
Feature: UI for Auth Validation Endpoint

sdewhitt:
*Issue #43*

Adds a management-console UI tab for the auth-validation endpoint
(`PUT /api/aws/auth/validate/:method`). Until now `web_ui/0` returned `[]`, so
the management plugin mounted no tab and the endpoint was API-only. This wires
up an admin-gated **Auth Validation** tab that drives the endpoint from the
browser, modeled on the `rabbitmq_queue_migration` management extension.

Scope is three files; no change to the endpoint's request pipeline (the one
`src/` edit is the `web_ui/0` return value).

### Changes

- **`src/aws_auth_validate_mgmt.erl`** -- `web_ui/0` now returns the JS asset,
  so the management plugin mounts the tab and serves this plugin's `priv/www`
  automatically (via `rabbit_mgmt_dispatcher`). No new routes needed.
- **`priv/www/js/aws_auth_validate.js`** (new) -- registers the Admin tab
  (`administrator` tag, matching the endpoint's default `required_user_tag`), a
  method selector (`ldap`/`http`/`oauth`/`tls`) with per-method field groups,
  and a hand-rolled request that maps the endpoint's fixed response categories
  (`input_invalid`, `connection_failed`, `tls_failed`, `auth_failed`,
  `config_conflict`, `authz_unverified`, `method_disabled`,
  `capacity_exhausted`, ...) to color-coded banners -- rather than routing
  through the console's CRUD helpers, which would surface a generic proxy error.
- **`priv/www/js/tmpl/aws_auth_validate.ejs`** (new) -- the form; field names
  mirror each backend's `allowed_fields/0` so the registry's `maps:with` filter
  keeps them.

### Security

Follows the endpoint's R6 invariant end to end:
- Secret-bearing fields (`password_arn`, `client_secret`, the mTLS cert/key
  ARNs, `access_token`) are sent in the JSON request body -- never in the
  URL/hash or Sammy route params (which land in browser history).
- Nothing is persisted to `localStorage` (the reference plugin persists form
  state; that was deliberately dropped so ARNs are never cached).
- Rendered text is HTML-escaped; only the backend's fixed, non-secret `message`
  is shown -- no target host, cert, or raw backend error is ever displayed.

The tab tag is a client-side visibility hint only; `is_authorized/2` remains the
authoritative gate.

### Testing

- Builds clean in the umbrella (`gmake`); assets confirmed served over HTTP on a
  running broker (`/api/extensions` lists the JS; the JS and EJS return 200).
- The endpoint itself was exercised end to end against the CDK LDAP test stack
  (a real simple-bind returns 204).

### DIFF
--- a/priv/www/js/aws_auth_validate.js
+++ b/priv/www/js/aws_auth_validate.js
@@ -397,12 +397,21 @@ function aws_auth_validate_merge_conf(method, existingText, body) {
     var spec = AWS_AUTH_VALIDATE_CONF[method];
 
     // Desired scalar key -> rendered value, for managed keys the form supplies.
+    // remove holds boolean keys the form does NOT supply: an unchecked checkbox
+    // is omitted from the body, and for booleans that means "off", so any stale
+    // line for it must be dropped rather than preserved. (Empty string/int fields
+    // are left absent from both maps and preserved verbatim below, so we never
+    // wipe config the operator did not touch.)
     var desired = {};
+    var remove = {};
     for (var key in spec.map) {
         if (!Object.prototype.hasOwnProperty.call(spec.map, key)) { continue; }
         var desc = spec.map[key];
         var v = aws_auth_validate_get_path(body, desc.path);
-        if (v === undefined || v === null || ('' + v).length === 0) { continue; }
+        if (v === undefined || v === null || ('' + v).length === 0) {
+            if (desc.type === 'bool') { remove[key.toLowerCase()] = true; }
+            continue;
+        }
         if (desc.type === 'bool') { v = v ? 'true' : 'false'; }
         desired[key.toLowerCase()] = {canonical: key, value: '' + v};
     }
@@ -461,8 +470,14 @@ function aws_auth_validate_merge_conf(method, existingText, body) {
             continue;
         }
 
-        // Any other line (unrelated config, another method's keys, a managed key
-        // the form left empty, assume_role) is preserved verbatim.
+        // Managed boolean key the form left unchecked ("off"): drop the stale
+        // line so unchecking (e.g. use_starttls) removes it from the config.
+        if (Object.prototype.hasOwnProperty.call(remove, k)) {
+            continue;
+        }
+
+        // Any other line (unrelated config, another method's keys, a managed
+        // string/int key the form left empty, assume_role) is preserved verbatim.
         out.push(line);
     }
 
@@ -558,33 +573,84 @@ function aws_auth_validate_config_status(text, cls) {
         .attr('class', 'argument' + (cls ? ' ' + cls : ''));
 }
 
-// Submit the current method. The rabbitmq.conf textarea is authoritative when it
-// holds content (an operator may have pasted or hand-edited config lines), so we
-// parse and send that; otherwise we build the body from the visible form and
-// reflect it back into the textarea as conf so what was validated is always
-// visible. Form-only fields (oauth access_token, tls target) have no conf key,
-// so they are always overlaid from the form. Bound as a delegated click so it
+// Submit the current method. The left-hand fields are the single source of
+// truth for validation; the rabbitmq.conf box is a paste/copy convenience that
+// never drives the request. This prevents the source-of-truth trap where the
+// fields are invalid but a differing pasted config looks valid -- we would
+// otherwise validate one thing while the operator reads another. If the box has
+// content that diverges from the fields, we block and point the operator at the
+// Parse / Update buttons to reconcile first. Bound as a delegated click so it
 // survives re-renders. Never a GET/route param -- the body may carry an ARN.
 $(document).on('click', '#aws-auth-validate-submit', function() {
     var method = $('#aws-auth-validate-method').val();
+    // Always build the request from the visible fields.
+    var body = aws_auth_validate_build_body(method);
+
     var raw = ('' + $('#aws-auth-validate-config').val()).trim();
-    var body;
     if (raw.length > 0) {
-        var parsed = aws_auth_validate_conf_to_body(method, raw);
-        body = parsed.body;
-        // Form-only fields never appear in conf; take them from the form.
-        aws_auth_validate_overlay_form_only(method, body);
+        // The box is non-empty: make sure it matches the fields before we run,
+        // so what is validated is exactly what is displayed. Overlay form-only
+        // fields (oauth access_token, tls target) onto the parsed body since
+        // they have no conf key and would otherwise read as a false divergence.
+        var pastedBody = aws_auth_validate_conf_to_body(method, raw).body;
+        aws_auth_validate_overlay_form_only(method, pastedBody);
+        if (aws_auth_validate_normalize_body(pastedBody) !==
+            aws_auth_validate_normalize_body(body)) {
+            aws_auth_validate_config_status(
+                'The config box differs from the fields. Validation runs on the ' +
+                'fields -- click "Parse into fields" to use the pasted config, or ' +
+                '"Update config from fields" to sync the box, then Validate again.',
+                'status-yellow');
+            return false;
+        }
     } else {
-        body = aws_auth_validate_build_body(method);
+        // Nothing pasted: reflect the fields into the box so the validated
+        // config is always visible and copyable.
         aws_auth_validate_set_config_text(aws_auth_validate_body_to_conf(method, body));
     }
+
     // Remember the exact conf text on screen so a 204 can mark it validated and
-    // the copy button can flag later edits. If the operator pasted conf we keep
-    // their text verbatim (comments and all); otherwise it is the generated conf.
+    // the copy button can flag later edits. It is now guaranteed consistent with
+    // the fields (verified above, or generated from them).
     var confText = '' + $('#aws-auth-validate-config').val();
+    // Clear any prior result and show a spinner immediately. A slow failure can
+    // take several seconds, during which a stale success banner would otherwise
+    // linger and mislead; the spinner makes "in progress" unambiguous.
+    aws_auth_validate_show_loading();
     aws_auth_validate_req(method, body, confText);
     return false;
 });
+
+// Stable stringification of a request body for divergence comparison: object
+// keys are sorted recursively so key order and whitespace never trigger a false
+// mismatch. Array order is preserved (the ldap servers list is meaningful).
+function aws_auth_validate_normalize_body(body) {
+    return JSON.stringify(aws_auth_validate_sort_keys(body));
+}
+function aws_auth_validate_sort_keys(v) {
+    if (Object.prototype.toString.call(v) === '[object Array]') {
+        return v.map(aws_auth_validate_sort_keys);
+    }
+    if (v && typeof v === 'object') {
+        var out = {};
+        Object.keys(v).sort().forEach(function(k) {
+            out[k] = aws_auth_validate_sort_keys(v[k]);
+        });
+        return out;
+    }
+    return v;
+}
+
+// Replace the result area with a spinner while a request is in flight. Cleared
+// by aws_auth_validate_render_result when the response (or transport error)
+// arrives.
+function aws_auth_validate_show_loading() {
+    $('#aws-auth-validate-result').html(
+        '<div class="status-grey" style="padding:8px;">' +
+        '<span class="aws-av-spinner"></span>' +
+        '<span class="argument">Validating&hellip;</span></div>'
+    );
+}
 
 // Parse the textarea rabbitmq.conf into the form fields. Lines that do not apply
 // to the selected method are ignored; aws.arns.assume_role_arn is skipped (read
--- a/priv/www/js/aws_auth_validate.js
+++ b/priv/www/js/aws_auth_validate.js
@@ -57,8 +57,10 @@ var AWS_AUTH_VALIDATE_CATEGORY = {
     tls_failed:         {cls: 'status-red',    text: 'TLS failed: handshake or certificate verification did not succeed.'},
     query_invalid:      {cls: 'status-red',    text: 'Query invalid: an authorization query could not be parsed.'},
     auth_failed:        {cls: 'status-yellow', text: 'Auth failed: the server was reached but did not authenticate/respond as expected.'},
-    config_conflict:    {cls: 'status-yellow', text: 'Config conflict: the supplied options are mutually inconsistent (for example an ARN is referenced with no assume_role configured).'},
-    authz_unverified:   {cls: 'status-yellow', text: 'Authorization unverified: a configured authorization check could not be confirmed.'},
+    token_invalid:      {cls: 'status-red',    text: 'Token invalid: the supplied access token failed signature verification, or no fetched JWKS key matched its key id -- the broker would also reject it.'},
+    token_expired:      {cls: 'status-yellow', text: 'Token expired: the supplied access token is past its exp (transient -- re-mint the token and retry).'},
+    config_conflict:    {cls: 'status-yellow', text: 'Config conflict: the supplied options are mutually inconsistent (for example an ARN is referenced with no assume_role configured, or an authorization check was requested but the oauth2 backend is not loaded on this broker).'},
+    authz_unverified:   {cls: 'status-yellow', text: 'Authorization unverified: the token is authentic but the requested permission could not be confirmed on the vhost/resource. Check scope_prefix, resource_server_id, additional_scopes_key, and scope_aliases.'},
     method_disabled:    {cls: 'status-yellow', text: 'Method disabled: enable it with aws.auth_validation.enabled_methods.<method> = true (every method is opt-in).'},
     unknown_method:     {cls: 'status-red',    text: 'Unknown method.'},
     insufficient_user_tag: {cls: 'status-red', text: 'Your user lacks the tag required to call this endpoint.'},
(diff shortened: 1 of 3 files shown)


priv/www/js/tmpl/aws_auth_validate.ejs
```diff
@@ -0,0 +1,205 @@
...
+
+    <!-- tls -->
+    <div id="aws-av-fields-tls" class="aws-av-method-fields" style="display:none;">
+      <table class="form">
+        <tr><th><label>Target:</label></th>
+          <td><input type="text" data-av-field="target" placeholder="listener" /></td></tr>
+      </table>
+    </div>
+
+    <!-- shared ssl_options (applies to ldap/http/oauth/tls that support TLS) -->
```
lukebakken: **HIGH.** This single shared `ssl_options` block is sent verbatim for every method, but each backend accepts a different nested-key set and rejects an unknown key **hard** as `input_invalid` (not a silent drop -- see `aws_auth_validate_ldap.erl` lines 378-380, with a comment saying exactly that):

- `ldap`: accepts `server_name_indication` (not `sni`) and has no `certfile_arn` / `keyfile_arn` (`aws_auth_validate_ldap.erl` lines 71-77)
- `tls`: no `sni`, no client cert/key pair (`aws_auth_validate_tls.erl` lines 60-66)
- `http` / `oauth`: `sni` plus the cert/key pair (`aws_auth_validate_oauth.erl` lines 131-139)

So filling SNI (or a client cert/key ARN) while `ldap` or `tls` is selected yields `input_invalid` with a generic banner. A bare LDAPS with only `cacertfile_arn` + `verify` still works, so this is scoped to the SNI/mTLS fields -- but those fields are unusable as presented. Consider per-method ssl field sets mirroring each backend's `SSL_OPTION_KEYS` (and the `sni` -> `server_name_indication` spelling for ldap). The build side is at `aws_auth_validate.js` line 104.

priv/www/js/aws_auth_validate.js
```diff
@@ -0,0 +1,824 @@
...
+// is collected from a small fixed set of sub-fields; ARN fields nest under it to
+// match the backend shape (ssl_options.cacertfile_arn etc.).
+function aws_auth_validate_build_body(method) {
+    var body = {};
+    var $form = $('#aws-auth-validate-form');
+
+    // Flat top-level fields per method. A checkbox contributes a boolean true
+    // only when checked (an unchecked box is omitted so the backend keeps its
+    // default); every other input contributes its non-empty value.
+    $form.find('[data-av-field]').each(function() {
```
lukebakken: **HIGH.** `aws_auth_validate_build_body` scans the entire form here (`$form.find('[data-av-field]')`), whereas `aws_auth_validate_populate_form` and `aws_auth_validate_overlay_form_only` scope to `$('#aws-av-fields-' + method)`. The method-selector change handler only hides the other groups; it never clears their inputs. So after filling one method and switching, the first method's stale hidden fields are merged into the new method's body. That then makes the submit-time divergence check (this side is form-wide, `conf_to_body` is method-scoped) permanently disagree, blocking Validate with "config box differs from the fields" -- and neither Parse nor Update (both method-scoped) can reconcile it. Scope this scan to `#aws-av-fields-<method>`, or clear the previous group on method change.

priv/www/js/aws_auth_validate.js
```diff
@@ -0,0 +1,824 @@
...
+    if (message) {
+        html += '<br/><span class="argument">' + fmt_escape_html(message) + '</span>';
+    }
+    html += ' <span class="argument">(HTTP ' + (req.status || 'n/a') + ')</span></div>';
+    $('#aws-auth-validate-result').html(html);
+}
+
+// Minimal HTML escaper (the console bundles fmt_escape_html in newer releases;
+// fall back to a local one so this works across broker versions).
+function fmt_escape_html(s) {
```
lukebakken: **HIGH.** This top-level `function fmt_escape_html` declaration installs itself as the global `window.fmt_escape_html`, overwriting the console's own (defined in `rabbitmq_management`'s `formatters.js` line 512). Extension JS loads after the core scripts, so the plugin's version wins process-wide once this tab is opened. Two consequences: (1) the guard `window.fmt_escape_html !== fmt_escape_html` compares the global against itself and is never true, so the "delegate to the console helper" branch is dead code; (2) every other console view that escapes multi-line text now runs this minimal escaper, which lacks the console version's `\n` -> `<br/>` handling. Rename to a plugin-scoped identifier (e.g. `aws_av_escape_html`) so it never shadows the global.

priv/www/js/aws_auth_validate.js
```diff
@@ -0,0 +1,824 @@
...
+    query_invalid:      {cls: 'status-red',    text: 'Query invalid: an authorization query could not be parsed.'},
+    auth_failed:        {cls: 'status-yellow', text: 'Auth failed: the server was reached but did not authenticate/respond as expected.'},
+    config_conflict:    {cls: 'status-yellow', text: 'Config conflict: the supplied options are mutually inconsistent (for example an ARN is referenced with no assume_role configured).'},
+    authz_unverified:   {cls: 'status-yellow', text: 'Authorization unverified: a configured authorization check could not be confirmed.'},
+    method_disabled:    {cls: 'status-yellow', text: 'Method disabled: enable it with aws.auth_validation.enabled_methods.<method> = true (every method is opt-in).'},
+    unknown_method:     {cls: 'status-red',    text: 'Unknown method.'},
+    insufficient_user_tag: {cls: 'status-red', text: 'Your user lacks the tag required to call this endpoint.'},
+    capacity_exhausted: {cls: 'status-yellow', text: 'Service at capacity or not ready. Try again shortly.'},
+    internal_error:     {cls: 'status-red',    text: 'Internal error during validation.'},
```
lukebakken, line 68: **Medium (blocking).** This category map has no entry for `token_expired` or `token_invalid`. The oauth backend returns both (`aws_auth_validate_oauth.erl` lines 669-737) and `status_for_category/1` maps them to 422 (`aws_auth_validate_mgmt.erl` lines 211-212). On a miss, `render_result` falls through to the generic "Unexpected response (422)" banner, defeating the transient (`token_expired`, just re-mint) vs config (`token_invalid`, bad signature/audience) distinction the backend deliberately exposes -- on the very oauth token path this UI is built to exercise. Add both keys here.

priv/www/js/aws_auth_validate.js
```diff
@@ -0,0 +1,824 @@
...
+    var $form = $('#aws-auth-validate-form');
+
+    // Flat top-level fields per method. A checkbox contributes a boolean true
+    // only when checked (an unchecked box is omitted so the backend keeps its
+    // default); every other input contributes its non-empty value.
+    $form.find('[data-av-field]').each(function() {
+        var el = $(this);
+        var key = el.attr('data-av-field');
+        if (el.is(':checkbox')) {
+            if (el.is(':checked')) { body[key] = true; }
```
lukebakken, line 98: **Medium.** Boolean asymmetry: an unchecked checkbox contributes no key here, but `aws_auth_validate_coerce('bool', 'false')` (line 306) returns `false` and `conf_to_body` stores `{key: false}`. So pasting a tutorial snippet with e.g. `auth_ldap.use_starttls = false` while leaving the (semantically equal) box unchecked makes the divergence check see `{use_starttls: false}` vs `{}` and blocks validation with "config box differs". The same asymmetry silently drops the explicit `false` on a Parse -> Update round-trip. Consider treating an unchecked box as an explicit `false`, or normalizing `false`-valued keys out of both sides before comparing.

priv/www/js/tmpl/aws_auth_validate.ejs
```diff
@@ -0,0 +1,205 @@
...
+    <div id="aws-av-fields-oauth" class="aws-av-method-fields" style="display:none;">
+      <table class="form">
+        <tr><th><label>JWKS URI:</label></th>
+          <td><input type="text" data-av-field="jwks_uri" placeholder="https://idp.example.com/.well-known/jwks.json" /></td></tr>
+        <tr><th><label>Issuer:</label></th>
+          <td><input type="text" data-av-field="issuer" placeholder="https://idp.example.com/" /></td></tr>
+        <tr><th><label>Resource server id:</label></th>
+          <td><input type="text" data-av-field="resource_server_id" placeholder="rabbitmq" /></td></tr>
+        <tr><th><label>Access token:</label></th>
+          <td><input type="text" data-av-field="access_token" placeholder="(optional) enables signature/exp/aud checks" /></td></tr>
```
lukebakken: **Low/medium (security posture).** `access_token` is a runtime bearer secret (deliberately kept out of conf text, clipboard, URL, and localStorage per R6), but it is rendered in a plain `type="text"` input, so it is shoulder-surfable and lingers visibly in the DOM for the session. It is never persisted, so severity is limited, but consider `type="password"` (or a masked input) to match the endpoint's no-secrets-exposed posture at the browser boundary.

priv/www/js/aws_auth_validate.js
```diff
@@ -0,0 +1,824 @@
...
+                // Drop the original server line (replaced by the block above).
+            } else {
+                out.push(line); // form has no servers -- leave existing untouched
+            }
+            continue;
+        }
+
+        // Managed scalar key the form supplies: update value in place, keeping
+        // the operator's original indentation and key spelling.
+        if (Object.prototype.hasOwnProperty.call(desired, k) && !seen[k]) {
```
lukebakken, line 590: **Low.** `merge_conf` updates only the first occurrence of a managed scalar key (the `!seen[k]` guard, then `seen[k] = true`). A second line with the same key falls through to the verbatim-preserve branch and keeps its stale value, so the merged text can carry two conflicting lines for one key. This is display/copy-only (the request is built from the fields), but since cuttlefish is last-assignment-wins, a config copied out of the box could silently override the freshly-updated value. Consider dropping later occurrences of an already-emitted managed key.

priv/www/js/aws_auth_validate.js
```diff
@@ -0,0 +1,824 @@
...
+    AWS_AUTH_VALIDATE_LAST_VALID = null;
+    aws_auth_validate_config_status('', '');
+    aws_auth_validate_show_template($('#aws-auth-validate-method').val());
+}
+
+// Admin-gated tab. The tag matches the endpoint's default required_user_tag
+// (administrator). This is a client-side visibility hint only; the endpoint's
+// is_authorized/2 remains the authoritative gate, and an operator who lowers
+// aws.auth_validation.required_user_tag does not change this hint.
+NAVIGATION['Admin'][0]['Auth Validation'] = ['#/auth-validate', "administrator"];
```
lukebakken, line 42: **Low (note).** The nav entry hardcodes the required tag as `administrator`, matching only the endpoint's default `required_user_tag`. The endpoint reads `aws.auth_validation_required_user_tag` and supports a lower tag; when an operator lowers it, the tab is hidden from the very (non-admin) audience meant to use it. Not a security issue -- `is_authorized/2` stays authoritative and the client hint is only stricter -- but a usability gap. The comment above already acknowledges the hint is static.

priv/www/js/aws_auth_validate.js
```diff
@@ -0,0 +1,824 @@
...
+// Issue the PUT and render the fixed category. Modeled on the reference plugin's
+// rqm_with_req: a hand-rolled request so we can inspect the status code and the
+// {error, message} body directly, rather than routing it through the management
+// plugin's CRUD helpers (which would surface a generic proxy error).
+function aws_auth_validate_req(method, body, confText) {
+    if (typeof has_auth_credentials === 'function' && !has_auth_credentials()) {
+        location.reload();
+        return;
+    }
+    var req = xmlHttpRequest();
```
lukebakken, line 995: **Low (note).** This request is created directly and is not registered in the console's `outstanding_reqs`, so unlike requests routed through the console's `with_req`, it is not aborted when the operator navigates away. A slow validation runs to completion (holding a semaphore slot the operator believes abandoned) and writes into a possibly-replaced DOM node. Minor, but a divergence from the reference plugins' request handling.

priv/www/js/tmpl/aws_auth_validate.ejs
```diff
@@ -0,0 +1,205 @@
...
+</div>
+
+<div class="section">
+  <h2>Result</h2>
+  <div id="aws-auth-validate-result">
+    <p class="argument">Submit a request to see the categorized result here.</p>
+  </div>
+</div>
+
+<script type="text/javascript">
```
lukebakken, line 383: **Low (note).** This inline `<script>` drives the on-render priming via jQuery `.html()` executing inline scripts. Under a hardened `content_security_policy` without `unsafe-inline` for `script-src`, the browser blocks it and the initial placeholder priming does not run (the state reset is redundant -- the sammy route already nulls `AWS_AUTH_VALIDATE_LAST_VALID`, the fields are static HTML, and the placeholder re-primes on any method change, so the tab still functions). This is **not** a deviation -- the core `feature-flags.ejs` uses the same pattern -- and CSP is off by default, so it is a low-priority observation rather than a blocker. Driving the priming from the sammy route / a post-render hook would avoid it.

priv/www/js/aws_auth_validate.js
```diff
@@ -0,0 +1,824 @@
...
+        // Server-list line: replace the whole run once with the form's servers.
+        if (spec.list && spec.list.re.test(k)) {
+            if (hasServers) {
+                if (!serverBlockEmitted) {
+                    for (var j = 0; j < serverLines.length; j++) { out.push(serverLines[j]); }
+                    serverBlockEmitted = true;
+                }
+                // Drop the original server line (replaced by the block above).
+            } else {
+                out.push(line); // form has no servers -- leave existing untouched
```
lukebakken: **Nit.** End-of-line comment. The repo `CLAUDE.md` "## Comments" section says: "Add comments above the line they are referring to, not at the end of the line". The rule's only example is Erlang (`%%`), so its applicability to JS is ambiguous, but if you want to be strict, move it above the line.

lukebakken (changes requested):
Review of the Auth Validation UI (the `web_ui/0` change plus the new JS and EJS). Method: eight independent finder angles, then adversarial verification of every candidate against source. Reviewed at 94d359b.

Checked and found sound (not defects): the asset-serving wiring (`rabbit_mgmt_dispatcher` serves this plugin's `priv/www`, filename matches), the admin gating, the OPTIONS/CORS preflight path, and the concern that the delegated `$(document).on(...)` handlers might double-bind on re-navigation (extension JS loads once at boot; `render()` only re-injects template HTML).

Blocking: the three HIGH inline comments plus the missing `token_expired` / `token_invalid` banners. The remaining inline comments are lower-severity notes and one nit.

priv/www/js/tmpl/aws_auth_validate.ejs
sdewhitt: Fixed in 034eeac. Split the shared block into per-method `ssl_options` sections matching each backend's allowed keys: ldap uses `cacertfile_arn`/`verify`/`server_name_indication` (no `sni`, no client cert); http and oauth add `certfile_arn`/`keyfile_arn`/`sni`; tls uses `cacertfile_arn`/`verify`/`fail_if_no_peer_cert`. The conf round-trip map was corrected to match, so Parse/Update no longer emit a key the backend rejects.

priv/www/js/aws_auth_validate.js
sdewhitt: Fixed in 034eeac. `build_body` (and `populate_form`'s clear/populate) now scope to `#aws-av-fields-<method>` instead of the whole form, so a hidden group's fields can no longer merge into the body.

priv/www/js/aws_auth_validate.js
sdewhitt: Fixed in 034eeac. Renamed to `aws_av_escape_html` so it no longer installs itself as `window.fmt_escape_html`; it prefers the console's global when present.

priv/www/js/aws_auth_validate.js
sdewhitt: Addressed -- `token_invalid` and `token_expired` entries are present in the category map (added with the OAuth authz work).

priv/www/js/aws_auth_validate.js
sdewhitt: Fixed in 034eeac. Added `strip_false_bools` to the divergence comparison only: an omitted boolean and an explicit `false` now compare equal (matching backend semantics). The request body is unchanged.

priv/www/js/tmpl/aws_auth_validate.ejs
sdewhitt: Fixed in 034eeac. `access_token` is now `type="password"` with `autocomplete="off"`.

priv/www/js/aws_auth_validate.js
sdewhitt: Fixed in 034eeac. A later duplicate of a managed key we already updated is now dropped, so cuttlefish last-wins can't resurrect a stale value. Unrelated duplicate lines are still preserved.

priv/www/js/aws_auth_validate.js
sdewhitt: Intentional: this is a client-side visibility hint only, and `is_authorized/2` is the authoritative gate. Hardcoding the default `administrator` tag is the conservative choice -- an operator who lowers `required_user_tag` still reaches the endpoint directly. Left as-is with the existing comment documenting it.

priv/www/js/aws_auth_validate.js
sdewhitt: Fixed in 034eeac. The request now registers with `outstanding_reqs` (guarded) and de-registers on completion, so navigate-away aborts it.

priv/www/js/tmpl/aws_auth_validate.ejs
sdewhitt: Left as-is: the core console primes its own views the same way via `.html()`-executed inline scripts, so this is consistent with the surrounding framework rather than a new deviation. Happy to revisit if a strict CSP is adopted plugin-wide.

priv/www/js/aws_auth_validate.js
sdewhitt: Fixed in 034eeac. Moved the comment above its line.

priv/www/js/aws_auth_validate.js
```diff
@@ -0,0 +1,1075 @@
...
+// (defined in the management plugin's formatters.js and relied on by other
+// views, e.g. for \n -> <br/> rendering). Prefer the console's global when it is
+// present so behavior matches the rest of the UI; fall back to this local
+// implementation on older brokers that do not bundle it.
+function aws_av_escape_html(s) {
+    if (typeof window.fmt_escape_html === 'function') {
+        return window.fmt_escape_html(s);
+    }
+    return ('' + s).replace(/&/g, '&amp;').replace(/</g, '&lt;')
+                   .replace(/>/g, '&gt;').replace(/"/g, '&quot;');
```
the-mikedavis: This is not currently necessary but for future-proofing we could add `.replace(/'/g, '&#39;')`, to be safe by default

the-mikedavis (approved):

PR #150, opened 2026-07-17
Auth Validation Endpoint SSRF/LDAP/audit hardening 

sdewhitt:
## Summary

Three focused hardening fixes to the auth-config validation endpoint, closing gaps found in review. No behavioral change to the happy path; each change tightens a security or observability edge.

## Changes

**1. SSRF: unwrap 6to4 (`2002::/16`) embedded IPv4 (`aws_auth_validate_ldap`)**

`is_private_ip6/1` already re-checked IPv4-mapped (`::ffff:a.b.c.d`), IPv4-compatible (`::a.b.c.d`), and NAT64 (`64:ff9b::a.b.c.d`) forms against the v4 private/infra ranges, but not 6to4. A 6to4-wrapped address such as `2002:a9fe:a9fe::` decodes to `169.254.169.254` and, on a host with a 6to4 route, would reach IMDS -- bypassing the LDAP SSRF classifier. This adds the 6to4 clause (embedded v4 lives in the 2nd/3rd 16-bit words), bringing the LDAP classifier back in step with the shared `aws_auth_validate_net:classify_ip/2`, which already handled 6to4.

**2. LDAP: server-side `timeLimit` on every post-bind search (`aws_auth_validate_ldap`)**

`eldap:open/2`'s timeout bounds each socket recv on the client side, but there was no protocol-level `timeLimit`. A server that accepts the bind and then dribbles a search response could hold a semaphore slot busy for the full connection timeout on each of a request's several sequential post-bind searches (dn_lookup, principal resolution, membership/existence). This passes a `{timeout, Seconds}` to each of those searches, derived from the configured connection timeout (ms -> s, rounded up, floored at 1s) so it tracks the operator's budget.

**3. Audit: log the authenticated management username (`aws_auth_validate_mgmt`)**

The endpoint is admin-gated and makes outbound connections to caller-chosen targets (an SSRF surface), so the acting principal must be attributable. Behind the management load balancer, source IP alone is often just the LB. Every audit path now includes the authenticated username. Only the username is logged -- never tags, target host, URL, or DN (preserves R4/R6). Also: an oversized body is now audited under its own `body_too_large` category instead of `input_invalid`, so the audit trail matches the returned status.

## Tests

Added to `aws_auth_validate_tests.erl`:
- `server_blocks_6to4_imds_test` / `server_allows_6to4_public_test` -- pre-connect `is_allowed_server/1` classifies 6to4-wrapped IMDS as blocked, 6to4-wrapped public as allowed (the embedded address decides, not the `2002` prefix).
- `peer_allowed_rebound_to_6to4_imds_blocked_test` -- the live-socket `peer_allowed/1` check catches a peer that rebound to 6to4-wrapped IMDS, matching the pre-connect check (defeats DNS rebinding on this form).

## Security invariants preserved
- No credential/DN/host leakage in logs -- only the username was added.
- SSRF guard remains resolve-and-pin; this only closes an embedded-v4 decode gap.
- No change to endpoint contract (methods, fields, error categories/statuses).

### DIFF
--- a/src/aws_auth_validate_ldap.erl
+++ b/src/aws_auth_validate_ldap.erl
@@ -559,17 +559,22 @@ is_private_ip(_) -> false.
 %%   fc00::/7       unique local addresses (ULA). Includes the IPv6 IMDS
 %%                  address fd00:ec2::254 -- the whole point of this block.
 %%   fe80::/10      link-local (spans fe80..febf, NOT just fe80)
-%% IPv4-mapped (::ffff:a.b.c.d), IPv4-compatible (::a.b.c.d), and NAT64
-%% (64:ff9b::a.b.c.d) addresses embed a v4 address in the low 32 bits; re-check
-%% those against the v4 ranges so e.g. ::ffff:169.254.169.254 (and the NAT64
-%% form 64:ff9b::169.254.169.254, which a host with a NAT64/DNS64 resolver would
-%% translate to IMDS) cannot reach IMDS.
+%% IPv4-mapped (::ffff:a.b.c.d), IPv4-compatible (::a.b.c.d), NAT64
+%% (64:ff9b::a.b.c.d), and 6to4 (2002:a.b.c.d::/16) addresses embed a v4 address;
+%% re-check those against the v4 ranges so e.g. ::ffff:169.254.169.254 (and the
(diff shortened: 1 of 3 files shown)


the-mikedavis (approved):
I don't have any comments. Nice work 🚀

PR #156, opened 2026-07-20
Fix: LDAP/UI divergence, shared module duplication, test coverage gaps

sdewhitt:
*Issues #151, #152, #153, #154, #155*

Addresses the post-merge review follow-ups filed as issues 151-155 against the
auth-validation endpoint. Each change is independent; grouped here as one PR
since they touch the same modules.

### #151 -- LDAPS hostname verification diverged from the broker (too lenient) [P-high bug]
The shared `apply_verify_default/2` unconditionally injected the RFC 6125 https
wildcard match fun on every `verify_peer` path. That is correct for the
http/oauth backends (IdP endpoints commonly present single-label wildcard certs,
e.g. Cognito), but it diverged from `rabbit_auth_backend_ldap`, which applies the
https fun **only** when `auth_ldap.ssl_options.hostname_verification = wildcard`
and otherwise uses OTP's stricter matching. A wildcard LDAPS/StartTLS cert the
live broker would reject at handshake passed the validator green.

- `aws_auth_validate_ssl`: add `apply_verify_default/3` with a `wildcard | strict`
  hostname-check mode. `/2` still defaults to `wildcard` (http/oauth unchanged).
- `aws_auth_validate_ldap`: model a `hostname_verification` ssl_options key
  (enum `wildcard | none`, mirroring the broker; unset = strict), validated in the
  pure phase, and pass the resolved mode into `build_tls_opts/2`.
- Console UI: add the `hostname_verification` field to the ldap CONF map and
  template so the FE/BE contract stays in sync.

### #152 -- UI divergence check misses nested tls `fail_if_no_peer_cert = false` [P-medium bug]
`strip_false_bools` stripped only top-level `false` booleans, so a pasted
`ssl_options.fail_if_no_peer_cert = false` versus an unchecked box compared
unequal and wrongly blocked submit with "The config box differs from the fields."

- `aws_auth_validate.js`: `strip_false_bools` now recurses one level into
  `ssl_options` (and drops an emptied `ssl_options`) so the false-vs-absent
  asymmetry is normalized. A pasted `true` vs an unchecked box still diverges.

### #153 -- LDAP SSRF IPv6 classifier duplicated the shared net module [P-medium refactor]
The embedded-v4 unwrapping was hand-copied between `aws_auth_validate_ldap`'s
`is_private_ip6/1` and `aws_auth_validate_net`'s `classify_ip/2` ("kept in step by
hand"). No live drift, but a future encoding or segment-math fix applied to one
copy and not the other would reopen a v6-encoded IMDS SSRF path.

- `aws_auth_validate_net`: extract the policy-independent unwrapping into a shared
  `embedded_v4/1` and route `classify_ip/2` through it.
- `aws_auth_validate_ldap`: `is_private_ip6/1` now shares that one unwrapper while
  keeping its own stricter range policy (LDAP denies all RFC1918/CGNAT).

### #154 -- test-coverage gaps in the new security code [P-medium]
Added unit tests (and the needed `-ifdef(TEST)` exports) for:
`apply_verify_default/3` strict-vs-wildcard, the shared `embedded_v4/1` matrix,
LDAP `hostname_verification` wiring, 6to4 `is_private_ip6`,
`search_time_limit_seconds` units/round-up/floor, `sanitize_log`
CR/LF/C0/DEL/UTF-8-preserved, `username/1` happy path + fallback, and the
`body_too_large` status mapping.

### #155 -- minor follow-ups [P-low]
- `aws_auth_validate_mgmt`: add `body_too_large` to `status_for_category/1` so a
  future refactor routing it through `respond/3` maps to 400, not 500.
- `aws_auth_validate_ldap`: soften the `search_time_limit_seconds` comment -- the
  LDAP protocol timeLimit is server-cooperative, not a hard DoS cap.
- `aws_auth_validate_ssl`: document the intended OS-trust-store fallback on the
  explicit-`verify_peer` no-anchor path.
- Console UI: note the byte-identical http/oauth ssl_options blocks must be
  hand-synced, clarify the symmetric fields-side `strip_false_bools` call, and
  trim the verbose `build_body` comment.

### Testing
`gmake eunit` in the umbrella (`deps/aws`): full suite passes (706 tests before
the new additions; the ~28 new test functions execute and pass). `erlfmt -c` is
clean on all changed `.erl` files.

### DIFF
--- a/priv/www/js/aws_auth_validate.js
+++ b/priv/www/js/aws_auth_validate.js
@@ -53,6 +53,7 @@ var AWS_AUTH_VALIDATE_METHODS = ['ldap', 'http', 'oauth', 'tls'];
 var AWS_AUTH_VALIDATE_CATEGORY = {
     success:            {cls: 'status-green',  text: 'Validation succeeded (204).'},
     input_invalid:      {cls: 'status-red',    text: 'Input invalid: the request failed pure validation before any connection.'},
+    body_too_large:     {cls: 'status-red',    text: 'Request body too large: reduce the config size and retry.'},
     connection_failed:  {cls: 'status-red',    text: 'Connection failed: the target could not be reached.'},
     tls_failed:         {cls: 'status-red',    text: 'TLS failed: handshake or certificate verification did not succeed.'},
     query_invalid:      {cls: 'status-red',    text: 'Query invalid: an authorization query could not be parsed.'},
@@ -338,11 +339,18 @@ var AWS_AUTH_VALIDATE_CONF = {
             'auth_http.resource_path':                          {path: ['resource_path'], type: 'string', ex: 'https://auth.example.com/auth/resource'},
             'auth_http.topic_path':                             {path: ['topic_path'], type: 'string', ex: 'https://auth.example.com/auth/topic'},
             'auth_http.http_method':                            {path: ['http_method'], type: 'string', ex: 'post'},
-            // The four ssl_options entries below are identical in shape to the
-            // oauth ssl_options entries; keep the two hand-synced on any change
-            // (only the auth_http/auth_oauth2 conf-key prefix differs).
+            // The ssl_options entries below match the oauth ssl_options entries
+            // key-for-key; keep the two hand-synced on any change. Only the
+            // conf-key PREFIX differs: http uses auth_http.ssl_options.*, while
+            // oauth uses auth_oauth2.ssl_options.* EXCEPT hostname_verification,
+            // which the broker spells auth_oauth2.https.hostname_verification.
             'auth_http.ssl_options.verify':                     {path: ['ssl_options', 'verify'], type: 'string', ex: 'verify_peer'},
             'auth_http.ssl_options.sni':                        {path: ['ssl_options', 'sni'], type: 'string', ex: 'auth.example.com'},
+            // Mirrors the broker's auth_http.ssl_options.hostname_verification
+            // (wildcard|none); unset means strict OTP matching. Modeled so the
+            // validator matches the broker's hostname check rather than always
+            // using the lenient wildcard fun.
+            'auth_http.ssl_options.hostname_verification':      {path: ['ssl_options', 'hostname_verification'], type: 'string', ex: 'wildcard'},
             'aws.arns.auth_http.ssl_options.cacertfile':        {path: ['ssl_options', 'cacertfile_arn'], type: 'string', ex: 'arn:aws:s3:::my-bucket/ca-cert.pem'},
             'aws.arns.auth_http.ssl_options.certfile':          {path: ['ssl_options', 'certfile_arn'], type: 'string', ex: 'arn:aws:s3:::my-bucket/client-cert.pem'},
             'aws.arns.auth_http.ssl_options.keyfile':           {path: ['ssl_options', 'keyfile_arn'], type: 'string', ex: 'arn:aws:secretsmanager:us-west-2:111122223333:secret:RabbitMqClientKey'}
@@ -360,6 +368,10 @@ var AWS_AUTH_VALIDATE_CONF = {
             'auth_oauth2.scope_pattern_syntax':                 {path: ['scope_pattern_syntax'], type: 'string', ex: 'wildcard'},
             'auth_oauth2.ssl_options.verify':                   {path: ['ssl_options', 'verify'], type: 'string', ex: 'verify_peer'},
             'auth_oauth2.ssl_options.sni':                      {path: ['ssl_options', 'sni'], type: 'string', ex: 'idp.example.com'},
+            // The broker spells this one auth_oauth2.https.hostname_verification
+            // (NOT under ssl_options), but it maps to the same request field
+            // (wildcard|none); unset means strict OTP matching.
+            'auth_oauth2.https.hostname_verification':          {path: ['ssl_options', 'hostname_verification'], type: 'string', ex: 'wildcard'},
             'aws.arns.auth_oauth2.ssl_options.cacertfile':      {path: ['ssl_options', 'cacertfile_arn'], type: 'string', ex: 'arn:aws:s3:::my-bucket/ca-cert.pem'},
             'aws.arns.auth_oauth2.ssl_options.certfile':        {path: ['ssl_options', 'certfile_arn'], type: 'string', ex: 'arn:aws:s3:::my-bucket/client-cert.pem'},
             'aws.arns.auth_oauth2.ssl_options.keyfile':         {path: ['ssl_options', 'keyfile_arn'], type: 'string', ex: 'arn:aws:secretsmanager:us-west-2:111122223333:secret:RabbitMqClientKey'}
--- a/priv/www/js/aws_auth_validate.js
+++ b/priv/www/js/aws_auth_validate.js
@@ -70,20 +70,13 @@ var AWS_AUTH_VALIDATE_CATEGORY = {
 };
 
 // Read the visible method form into a JSON request body. Only non-empty fields
-// are included so an omitted optional field keeps the backend default. ssl_options
-// is collected from the selected method's own ssl sub-fields; ARN fields nest
-// under it to match the backend shape (ssl_options.cacertfile_arn etc.).
+// are included so an omitted optional field keeps the backend default; ARN
+// fields nest under ssl_options to match the backend shape.
 //
-// The scan is scoped to the SELECTED method's field container
-// (#aws-av-fields-<method>), never the whole form. Each method group carries its
-// own data-av-field / data-av-ssl inputs (the ssl_options key set differs per
-// backend -- e.g. ldap wants server_name_indication and has no client cert,
-// while http/oauth want sni + certfile/keyfile), and switching methods only
-// HIDES the other groups. A whole-form scan would therefore merge a previously-
-// selected method's (now hidden) values into this body -- producing a request
-// with foreign keys the backend rejects as input_invalid, and a permanent
-// divergence from the conf box. Scoping to the active group is what keeps the
-// body method-pure.
+// The scan is scoped to the SELECTED method's container (#aws-av-fields-<method>),
+// never the whole form: each method's ssl_options key set differs, and switching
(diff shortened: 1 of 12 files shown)


the-mikedavis (approved):

src/aws_auth_validate_ssl.erl
```diff
@@ -377,36 +378,58 @@ translate_ssl_opts(Map, SniKey) ->
...
-%% false-negative this endpoint exists to avoid.
--spec apply_verify_default(list(), boolean()) ->
-    {ok, list()} | {error, aws_auth_validate_backend:error_category(), binary()}.
+%% The /2 form applies the RFC 6125 https hostname-match fun on every verify_peer
+%% path (see ensure_hostname_check/1). This is correct for the http/oauth
+%% backends: their IdP endpoints commonly present single-label wildcard certs
+%% (e.g. Cognito's `foo.auth.<region>.amazoncognito.com' against
+%% `*.auth.<region>.amazoncognito.com') that every mainstream HTTPS client
+%% accepts, so omitting the fun would report tls_failed for valid endpoints.
 apply_verify_default(Opts, VerifyExplicit) ->
```
lukebakken, line 405: **HIGH (blocking).** The #151 fix is correctly generalized into `apply_verify_default/3` with a `strict | wildcard` mode, but only the LDAP backend was switched to it. This `/2` form hardcodes `wildcard`, and both `aws_auth_validate_http.erl:772` and `aws_auth_validate_oauth.erl:1181` still call it -- so http and oauth unconditionally inject the RFC 6125 https wildcard match fun on every `verify_peer` path.

The real broker gates that match fun behind a strict-by-default opt-in for **all three** backends, not just LDAP:
- http: `rabbit_auth_backend_http.erl:262-270` adds it only on `{ok, wildcard}` for `ssl_hostname_verification`; the unset (`_`) branch is strict
- oauth: `oauth2_client.erl:447-453` -- `maps:get(hostname_verification, Map, none)`, so unset = `none` = strict
- ldap: now fixed here to default `strict`

So the exact too-lenient divergence #151 closed for LDAP is still live for http/oauth: a wildcard or hostname-mismatched IdP / HTTP-auth server cert passes the validator green, but a default-configured broker rejects it at handshake. The comment at lines 381-386 asserting wildcard "is correct for the http/oauth backends" is contradicted by the broker source above -- their default is strict, and wildcard is opt-in.

Suggested fix (mirroring what was just done for LDAP): model a `hostname_verification` key (enum `wildcard | none`, default strict) in the http and oauth `SSL_OPTION_KEYS` + pure-phase validation, compute the mode, and call `apply_verify_default/3` at both call sites; add the field to the CONF map and template for both methods; and correct the 381-386 comment. If the hardcoded wildcard was in fact deliberate for these backends, could you note the rationale, since the broker source points the other way?

priv/www/js/aws_auth_validate.js
```diff
@@ -70,20 +70,13 @@ var AWS_AUTH_VALIDATE_CATEGORY = {
 };
```
lukebakken, line 71: **Low (UX).** `AWS_AUTH_VALIDATE_CATEGORY` (the map just above, lines 53-69) has no `body_too_large` entry, but the backend returns `{error: "body_too_large"}` at HTTP 400 (`aws_auth_validate_mgmt.erl:107-109`). On an oversized body the render path misses the lookup and falls through to the generic "Unexpected response (400)." banner instead of a category-specific message. This is pre-existing, but this PR promotes `body_too_large` to a first-class category (adds `status_for_category(body_too_large) -> 400` plus a test), and the map's own comment says it is "kept in one place so the .ejs and the banner stay in sync" -- now not quite true for this one category. The operator still sees the backend's fixed `message` string, so severity is cosmetic. One-line fix: add `body_too_large: {cls: 'status-red', text: 'Request body too large.'}`.

src/aws_auth_validate_ssl.erl
```diff
@@ -424,6 +447,13 @@ ensure_hostname_check(Opts) ->
 %% Return {ok, OptsWithAnchor} when a trust anchor is present or can be sourced
 %% from the OS store, else `none'. (trust_source/1 is the historical LDAP name
 %% for the same behaviour; kept as an alias for that backend's call site.)
+%%
+%% Note (intended): when no cacertfile_arn is supplied, verify_peer falls back to
```
lukebakken, line 480: **Discussion (not blocking).** This "Note (intended)" correctly documents the #155 item: on the no-`cacertfile_arn` + `verify_peer` path, `ensure_trust_anchor/1` sources the validator host's OS CA store (`public_key:cacerts_get()`), so the validator's trust store -- not the broker's configured anchor -- decides the verdict. That is a real parity gap (a cert chaining to a public CA the validator trusts but the broker is not configured to trust reports green, then fails in production), and it points in the *opposite* direction from the LDAP hostname-matching tightening in this same PR. Documenting it satisfies #155's "confirm this is intended" ask, so no change required -- but now that hostname matching was aligned to broker-parity, is leaving this trust-anchor leniency the consistent call? Fine to keep as-is if the convenience (probing public-CA endpoints without a redundant ARN) is the intended trade-off.

priv/www/js/aws_auth_validate.js
```diff
@@ -689,21 +690,27 @@ function aws_auth_validate_strip_nonconf(method, body) {
...
+// recursing it would read as a false divergence. An ssl_options object emptied by
+// stripping is dropped so it matches a fields side that never set it. Non-boolean
+// values and `true' are left as-is, so this never masks a real difference.
 function aws_auth_validate_strip_false_bools(body) {
     if (!body || typeof body !== 'object') { return body; }
     var out = {};
     for (var k in body) {
         if (!Object.prototype.hasOwnProperty.call(body, k)) { continue; }
         if (body[k] === false) { continue; }
+        if (k === 'ssl_options' && body[k] && typeof body[k] === 'object') {
```
lukebakken, line 721: **Low (note, no change needed).** The recursion is gated on the literal key name `'ssl_options'` rather than a structural nested-object test. It is correct today -- the other nested structures `build_body` emits (`queries`, `scope_aliases`, `authz_check`) carry no booleans and are removed by `strip_nonconf` before this runs -- so this is only a future-drift note: if a new conf-representable nested object ever carries a boolean, the omit-vs-false divergence would silently reappear for it. A naive structural rewrite is *worse* here (in JS arrays are `typeof === 'object'`, so `servers` / `scope_aliases` would get wrongly descended, and the empty-drop semantics are ssl_options-specific), so the targeted key match is a defensible choice. Flagging only so the coupling is a known trade-off.

priv/www/js/tmpl/aws_auth_validate.ejs
```diff
@@ -204,7 +209,10 @@ SPDX-License-Identifier: Apache-2.0
             <option value="get">get</option><option value="post">post</option></select></td></tr>
       </table>
 
-      <!-- http ssl_options: cacertfile_arn, certfile_arn, keyfile_arn, verify, sni. -->
+      <!-- http ssl_options: cacertfile_arn, certfile_arn, keyfile_arn, verify, sni.
+           NOTE: this block is byte-identical to the oauth ssl_options block below;
```
lukebakken: **Low.** These NOTE comments (here and the mirror on the oauth block below, plus the CONF-map counterpart at `aws_auth_validate.js:342`) formalize a manual-duplication contract -- the http and oauth `ssl_options` blocks are byte-identical and must be "hand-synced on any ssl change". A single logical ssl change now takes four coordinated edits (two template blocks + two CONF entries) with nothing enforcing the sync. Notably the blocking #151 fix above is exactly such a change (adding a `hostname_verification` field to both http and oauth), so this is about to bite. Consider driving both method blocks from one shared ssl-field descriptor instead; folding that into the #151 fix would resolve both at once.

lukebakken (changes requested):
Review of the follow-up fixes for #151-#155. Method: eight independent finder angles, then adversarial verification of every load-bearing candidate against source (including the real broker backends). Reviewed at b9ef11f.

**Resolved and verified sound:**
- **#152** (nested tls `fail_if_no_peer_cert` divergence): `strip_false_bools` now recurses into `ssl_options`, symmetric on both comparison sides, with tests
- **#153** (SSRF classifier duplication): the embedded-v4 unwrapper is now the single shared `aws_auth_validate_net:embedded_v4/1`, used by both `classify_ip/2` and `is_private_ip6/1`; verified behaviour-preserving across all four v6-carrying encodings (IPv4-mapped, IPv4-compatible, NAT64, 6to4), LDAP's stricter range policy intact
- **#154** (test-coverage gaps): all six named gaps now have real, non-vacuous tests (the round-up test uses 4001ms to genuinely pin the ceil, the 6to4 tests assert opposite verdicts on IMDS vs public, and the no-anchor tests mock an empty OS store for determinism)

**Blocking:** one HIGH inline comment -- the #151 fix landed for LDAP only; http and oauth are still on the too-lenient path. The rest are lower-severity inline notes and one discussion point.

lukebakken (approved):
Re-reviewed the follow-up commit 68c5948. The blocking #151 item is resolved at the right altitude: http and oauth now call apply_verify_default/3 with a shared hostname_check_mode/1 (strict by default, wildcard opt-in), matching the broker's gate for all three backends. The hostname_verification value is enum-validated in the pure phase, and it does not leak through translate_ssl_opts into the actual ssl option list. body_too_large banner and the http/oauth CONF/template sync are in place. LGTM.

src/aws_auth_validate_ssl.erl
sdewhitt: Fixed in 68c5948. Rather than hardcode wildcard, I generalized the LDAP approach:
- Added a shared aws_auth_validate_ssl:hostname_check_mode/1 (unset/none → strict, wildcard → opt-in) plus hostname_verification value validation (enum wildcard|none) in the shared module.
- aws_auth_validate_http and aws_auth_validate_oauth now model hostname_verification in SSL_OPTION_KEYS + the reasons map, compute the mode, and call apply_verify_default/3.
- aws_auth_validate_ldap drops its private hostname_check_mode/1 in favour of the shared one (no behaviour change).
- Corrected the 381-386 comment. it now states the broker default is strict for all three backends and wildcard is opt-in, matching the sources you cited.

PR #159, opened 2026-07-20
Hardening/auth validation parity followups

sdewhitt:
*Issue #, if available:* Closes #157, #158. Stacks additional hardening follow-ups with no individual issues.

Post-merge hardening follow-ups for the auth-validation endpoint, in the same vein as # 151-155. 

### 1. LDAP hardening
- **shared range-matching:** express the LDAP SSRF range policy as `?LDAP_DENIED_V4_CIDRS` / `?LDAP_DENIED_V6_CIDRS` consumed by the shared `aws_auth_validate_net:in_any_cidr/2`, replacing the hand-coded guard clauses. LDAP keeps its own broader list (all RFC1918/CGNAT/reserved, not just broker infra), but both backends now enumerate ranges in ONE vocabulary, so a future range or segment-math fix cannot silently apply to only one classifier. Mirrors the #153 `embedded_v4` unification, one layer up.
- **port parity:** default an omitted `port` to 389, matching `rabbit_auth_backend_ldap`'s application-env default (the broker does not bump to 636 for TLS). A config the live broker would accept no longer fails `input_invalid`.
- **loopback-only test flag:** the test-only `auth_validation_allow_private_networks` flag now relaxes ONLY loopback (via `is_denied_ip/1`, mirroring `aws_auth_validate_net:classify_denied/1`) instead of fully bypassing both SSRF checks. IMDS/link-local/RFC1918/CGNAT stay denied even under the flag, so the LDAP backend no longer gets a broader test-mode bypass than http/oauth. Both the pre-connect and post-connect peer checks always run.

### 2. FE/BE category drift + oauth jwks_uri query parity (#157, #158)
- **#157:** add `body_too_large` to `AWS_AUTH_VALIDATE_CATEGORY`. The backend returns `{error:"body_too_large"}` at HTTP 400, but the UI status map never had the matching key, so an oversized body fell through to the generic "Unexpected response (400)" banner. It was the sole backend category missing from the map.
- **#158:** allow a pre-existing query string on `jwks_uri` (both directly-supplied and OIDC-discovered). `jwks_uri` is fetched VERBATIM -- the live broker's `uaa_jwks:get/2` fetches it as-is -- so rejecting a query string `input_invalid` diverged too-strict from the broker. `issuer` keeps rejecting a query (OIDC discovery appends the well-known path). Threaded via a `QueryPolicy` (`allow_query | reject_query`) arg on `parse_url/2`.

### 3. Build fix
- Drop the now-unused `parse_url/1` wrapper (both call sites moved to `parse_url/2`). It only survived the local eunit build because it was in the `-ifdef(TEST)` export list; the production `make app` build (warnings-as-errors, no TEST exports) flagged it as unused. Test export updated to `parse_url/2`.

### Testing
- `erlfmt -c` clean on all `.erl` changes.
- Production `make app` build (warnings-as-errors) passes.
- Full clean eunit run green.
- Verified this branch merges cleanly on top of #156 (no conflicts; 728 tests pass on the combined tree).

### DIFF
--- a/priv/www/js/aws_auth_validate.js
+++ b/priv/www/js/aws_auth_validate.js
@@ -53,6 +53,7 @@ var AWS_AUTH_VALIDATE_METHODS = ['ldap', 'http', 'oauth', 'tls'];
 var AWS_AUTH_VALIDATE_CATEGORY = {
     success:            {cls: 'status-green',  text: 'Validation succeeded (204).'},
     input_invalid:      {cls: 'status-red',    text: 'Input invalid: the request failed pure validation before any connection.'},
+    body_too_large:     {cls: 'status-red',    text: 'Request body too large: reduce the config size and retry.'},
     connection_failed:  {cls: 'status-red',    text: 'Connection failed: the target could not be reached.'},
     tls_failed:         {cls: 'status-red',    text: 'TLS failed: handshake or certificate verification did not succeed.'},
     query_invalid:      {cls: 'status-red',    text: 'Query invalid: an authorization query could not be parsed.'},
--- a/src/aws_auth_validate_http.erl
+++ b/src/aws_auth_validate_http.erl
@@ -365,44 +365,14 @@ collect_paths([Key | Rest], Body, Acc, Paths) ->
             {error, input_invalid, ?REASON_BAD_PATH_VALUE}
     end.
 
-%% Parse + minimally validate an http(s) URL. Returns a normalized
-%% representation the probe and the SSRF guard can both use. Kept deliberately
-%% small here; richer parsing belongs with the guard work.
+%% Parse + minimally validate an http(s) URL via the shared
+%% aws_auth_validate_net:parse_url/2. The broker's auth_http.*_path config is
+%% query- and fragment-less and the probe appends its own ?username= params, so
+%% a pre-existing query (double-? mangling) or #fragment (probe's query dropped
+%% after the fragment) is rejected -- both are the parser's default. userinfo and
+%% out-of-range ports are always rejected.
 parse_url(Bin) when is_binary(Bin) ->
-    Str = binary_to_list(Bin),
-    case uri_string:parse(Str) of
-        #{scheme := Scheme, host := Host} = Parsed when
-            Host =/= [], (Scheme =:= "http" orelse Scheme =:= "https")
-        ->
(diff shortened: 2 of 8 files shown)


lukebakken (commented):
Pushed 8a06b96, which addresses the review findings on this PR.

## Overview

@sdewhitt this consolidates the URL parsing and SSRF address classification that the http, oauth, and ldap auth-validation backends had each hand-rolled. The three copies had already drifted, so this collapses them onto shared code and fixes a latent bug found along the way.

## What changed

- Moved URL parsing into a shared `aws_auth_validate_net:parse_url/2`, driven by a `url_opts()` map (scheme allowlist plus per-field query/fragment policy, both defaulting to reject). http and oauth now delegate to it
- Switched ldap to classify addresses through the shared `classify_ip/2` with its own stricter denylist, instead of a private reimplementation. This removes ldap's `is_private_ip/1`, `is_private_ip6/1`, `is_denied_ip6/1`, `relaxed_loopback/1`, and the duplicated CIDR macros. The v6-embedded-v4 unwrap and the loopback-only test relaxation now live in exactly one place, so ldap can no longer drift from http/oauth on how e.g. `::ffff:127.0.0.1` is handled
- Added a `cidr_policy()` type so ldap can pass a denylist-only map without dummy scheme/reason keys, and widened `classify_ip/2` to accept it. This keeps the shared call dialyzer-clean

## Bug fixed along the way

The oauth fragment policy is now tied to the same per-field decision as its query policy. The `issuer` URL has the OIDC well-known path appended, so a pre-existing `#fragment` mangles the discovery request exactly as a query string does (the path lands after the `#` and httpc drops it). `jwks_uri` is fetched verbatim, so a fragment there is harmless. Previously the parser never inspected the fragment at all, so an `issuer` with a fragment was accepted and its discovery request silently misfired. This is a latent pre-existing bug, not a regression from this refactor; tracked as #162.

## Tests and checks

- Added regression tests pinning issuer-fragment rejection, jwks_uri-fragment acceptance, and the `::ffff:127.0.0.1` test-flag relaxation
- Full eunit suite passes (732 tests)
- `make app` clean with warnings-as-errors
- Confirmed via a stash-and-rebuild baseline that these changes add zero new dialyzer warnings (the existing backlog is tracked in #8)

lukebakken (approved):

PR #160, opened 2026-07-21
OAuth Authz validation parity test enhancements

sdewhitt:
Lands after #159

## What this adds

A parity-pin test for the OAuth authz validation endpoint that captures the current upstream behavior around dotted `additional_scopes_key` values (rabbitmq-server 16947), plus a small comment correction on the existing over-trust caveat.

rabbitmq-server 16947 documents that the OAuth backend resolves `additional_scopes_key` via `rabbit_auth_backend_oauth2:split_path/1`, which splits the key on every `.`:

```erlang
split_path(Path) -> binary:split(Path, <<".">>, [global, trim_all]).
```

A flat claim key that itself contains dots — e.g. an STS/OIDC key like `https://sts.amazonaws.com/tags` — is therefore split into a nested-map path (`["https://sts", "amazonaws", "com/tags"]`), the flat claim is never found, and its scopes are not extracted. The validation endpoint replays the broker's real scope-resolution code, so it faithfully reproduces this behavior.

## Changes

- **`test/aws_auth_validate_oauth_tests.erl`** — adds `authz_dotted_additional_scopes_key_parity_pin_test_`:
  - Mints a token carrying scopes under a flat dotted claim key (`https://sts.amazonaws.com/tags`), points `additional_scopes_key` at that key, and runs it through the real upstream `normalize_token_scope` / `get_expanded_scopes` / `resource_access` path (not mocks).
  - Asserts the current behavior: `{error, authz_unverified, _}` with the "no scopes for this resource_server" message, because `split_path/1` mis-splits the dotted key.
  - Gated by `maybe_skip_authz/1`, so it skips cleanly on a pre-floor broker series that lacks the arity-4 scope API rather than failing.
  - Pins upstream behavior: if `split_path` is later changed (or a dedicated flat-key accessor is added), this test flips and forces a coordinated broker-dependency bump — surfacing the drift as a test failure the moment the dependency changes.

- **`src/aws_auth_validate_oauth.erl`** — corrects the module's over-trust caveat comment to reflect that scope authorization can now be checked via the optional `authz_check` block (delegated to `aws_auth_validate_oauth_authz`), while noting the documented resource-server-field overlay parity gap.

## Why pin rather than fix

Fixing the dotted-key handling locally in the validation endpoint would make it report a grant the live broker would still refuse — a false pass, and exactly the parity regression the endpoint exists to avoid. The correct behavior is to mirror the current upstream behavior and pin it, so the endpoint and the real broker always agree. The follow-on escaped-dot support (once rabbitmq-server 16985 lands upstream) is tracked separately in #161, which builds on this PR.

## Verification

- `gmake eunit t=aws_auth_validate_oauth_tests` — all tests pass; the new pin test's assertions confirmed executing (not skipped) with the OAuth backend on the code path.
- `erlfmt -c` clean.

### DIFF
--- a/test/aws_auth_validate_oauth_tests.erl
+++ b/test/aws_auth_validate_oauth_tests.erl
@@ -1032,6 +1032,66 @@ authz_iam_scope_alias_grants_access_test_() ->
         end)
     end}.
 
+%% -------- PARITY PIN: rabbitmq/rabbitmq-server discussion #16947 --------
+%%
+%% rabbit_auth_backend_oauth2 resolves additional_scopes_key via split_path/1:
+%%
+%%   split_path(Path) ->
+%%       binary:split(Path, <<".">>, [global, trim_all]).
+%%
+%% Flat claim keys containing dots -- typical for OIDC/STS-style URIs like
+%% <<"https://sts.amazonaws.com/tags">> -- are split into nested-map path
+%% segments (["https://sts", "amazonaws", "com/tags"]) and looked up as a
+%% deeply nested key in the flat claims map. The key is never found, so the
+%% scopes under it are never extracted.
+%%
+%% This test PINS the current (broken) behavior: the authz evaluator returns
+%% authz_unverified because it sees no effective scopes. When upstream fixes
+%% split_path (or adds a dedicated flat-key accessor for additional_scopes_key),
+%% this test MUST be revisited: the assertion should flip from authz_unverified
+%% to ok, and the change landed together with the broker dependency bump.
+%%
+%% This matches the project's parity stance: document upstream behavior, pin
+%% against drift, and surface regressions as test failures the moment the
+%% dependency changes.
+%% -------------------------------------------------------------------
+authz_dotted_additional_scopes_key_parity_pin_test_() ->
+    {setup, fun setup_httpc_mock/0, fun teardown_httpc_mock/1, fun(_) ->
+        maybe_skip_authz(fun() ->
+            #{jwk_pub := PubJwk, sign := Sign} = rsa_signer(<<"k1">>),
+            %% Token carries scopes under a dotted claim key (OIDC/STS-style URI).
+            %% split_path/1 will incorrectly split this on dots, breaking resolution.
+            Token = Sign(#{
+                <<"exp">> => future(),
+                <<"aud">> => <<"rabbitmq">>,
+                <<"https://sts.amazonaws.com/tags">> => [
+                    <<"rabbitmq.write:*/*">>, <<"rabbitmq.read:*/*">>
+                ]
+            }),
+            JwksBody = rabbit_json:encode(#{<<"keys">> => [PubJwk]}),
+            mock_httpc_response(200, JwksBody),
+            Body = (jwks_body())#{
+                <<"access_token">> => Token,
+                <<"resource_server_id">> => <<"rabbitmq">>,
+                <<"scope_prefix">> => <<"rabbitmq.">>,
+                %% Point additional_scopes_key at the dotted claim.
+                <<"additional_scopes_key">> => <<"https://sts.amazonaws.com/tags">>,
+                <<"authz_check">> => #{
+                    <<"resource">> => <<"my-queue">>,
+                    <<"permission">> => <<"write">>
+                }
+            },
+            Result = aws_auth_validate_oauth:validate(Body),
+            [
+                %% Current behavior: split_path splits the dotted key, lookup fails,
+                %% no scopes extracted -> authz_unverified. This will flip to ok once
+                %% upstream fixes the lookup.
+                ?_assertMatch({error, authz_unverified, _}, Result),
+                ?_assert(reason_contains(Result, "no scopes for this resource_server"))
+            ]
+        end)
+    end}.
+
 %% authz_check without an access_token is rejected in the pure phase.
 authz_check_without_token_rejected_test() ->
     Body = (jwks_body())#{
--- a/test/aws_auth_validate_oauth_tests.erl
+++ b/test/aws_auth_validate_oauth_tests.erl
@@ -1051,9 +1051,24 @@ authz_iam_scope_alias_grants_access_test_() ->
 %% this test MUST be revisited: the assertion should flip from authz_unverified
 %% to ok, and the change landed together with the broker dependency bump.
 %%
-%% This matches the project's parity stance: document upstream behavior, pin
-%% against drift, and surface regressions as test failures the moment the
-%% dependency changes.
+%% DORMANCY: the pin only RUNS once the broker ships the arity-4 scope API
+%% (resource_access/4). On an older dep, HAVE_OAUTH2_RESOURCE_SERVER is undefined,
+%% available/0 is false, and maybe_skip_authz/1 returns [] -- the body never
+%% executes. So a green suite does NOT mean "still broken as expected"; it may
+%% mean the pin was skipped. Concretely, if upstream backports only the
+%% split_path fix to a dep series WITHOUT the arity-4 API, this pin stays dormant
+%% and the regression slips through here. The pin fires only on a series with the
+%% full arity-4 API; treat the parity-bump revisit as the real backstop.
+%%
+%% FLIP SHAPE: the scope value below is deliberately `rabbitmq.'-prefixed. In the
+%% fixed backend, extract_scope_list_from_token_value/2 takes a LIST value
+%% VERBATIM (only a map value keyed by resource_server_id is prefix-injected), so
+%% after the scope_prefix "rabbitmq." strip this yields write:*/* and flips to ok.
+%% An unprefixed value here would be filtered to [] by the strip and never flip.
+%%
+%% This matches the project's parity stance: document upstream behavior and pin
+%% against drift, surfacing regressions as test failures once the dep ships the
+%% arity-4 API.
 %% -------------------------------------------------------------------
 authz_dotted_additional_scopes_key_parity_pin_test_() ->
     {setup, fun setup_httpc_mock/0, fun teardown_httpc_mock/1, fun(_) ->
--- a/test/aws_auth_validate_oauth_tests.erl
+++ b/test/aws_auth_validate_oauth_tests.erl
@@ -1051,6 +1051,14 @@ authz_iam_scope_alias_grants_access_test_() ->
 %% this test MUST be revisited: the assertion should flip from authz_unverified
 %% to ok, and the change landed together with the broker dependency bump.
 %%
+%% CONTROL: the authz_unverified assertion alone does not isolate the dotted-key
+%% cause -- authz_no_effective_scopes_after_prefix_test_ asserts the identical
+%% category and message for a wrong-prefix token, so ANY unresolved
+%% additional_scopes_key looks the same. The paired control case sends the SAME
+%% scopes under a NON-dotted key and asserts it reaches ok, making the dot the
+%% only variable: if the control ever stops granting, the breakage is in
+%% additional_scopes_key resolution generally, not in the dotted-key split.
+%%
 %% DORMANCY: the pin only RUNS once the broker ships the arity-4 scope API
 %% (resource_access/4). On an older dep, HAVE_OAUTH2_RESOURCE_SERVER is undefined,
 %% available/0 is false, and maybe_skip_authz/1 returns [] -- the body never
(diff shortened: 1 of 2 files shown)


test/aws_auth_validate_oauth_tests.erl
```diff
@@ -1032,6 +1032,66 @@ authz_iam_scope_alias_grants_access_test_() ->
...
+%% this test MUST be revisited: the assertion should flip from authz_unverified
+%% to ok, and the change landed together with the broker dependency bump.
+%%
+%% This matches the project's parity stance: document upstream behavior, pin
+%% against drift, and surface regressions as test failures the moment the
+%% dependency changes.
+%% -------------------------------------------------------------------
+authz_dotted_additional_scopes_key_parity_pin_test_() ->
+    {setup, fun setup_httpc_mock/0, fun teardown_httpc_mock/1, fun(_) ->
+        maybe_skip_authz(fun() ->
```
lukebakken, line 1083: This pin is unconditionally skipped on the branch it lives on, which undermines the comment's promise (lines 1054-1056) to "surface regressions as test failures the moment the dependency changes."

On this checkout the resolved `rabbitmq_auth_backend_oauth2/include/oauth2.hrl` has no `scope_pattern_syntax` field and `rabbit_oauth2_scope:resource_access` is arity-3, so the Makefile does not define `HAVE_OAUTH2_RESOURCE_SERVER`. `aws_auth_validate_oauth_authz` then compiles its `-else` branch where `available/0` returns `false`, so `maybe_skip_authz/1` returns `[]` and this body never runs.

Concrete drift that slips through: if upstream backports only the `split_path/1` fix for #16947 to a v4.3.x-series oauth2 backend without also landing the arity-4 scope API, `available/0` stays `false`, the pin stays skipped, and the regression it targets goes unnoticed here. The pin fires only on a broker series that ships the full arity-4 API. Worth a note in the comment that the pin is dormant until the arity-4 scope API is present, so a reader does not assume green means "still broken as expected."

test/aws_auth_validate_oauth_tests.erl
```diff
@@ -1032,6 +1032,66 @@ authz_iam_scope_alias_grants_access_test_() ->
...
+                    <<"permission">> => <<"write">>
+                }
+            },
+            Result = aws_auth_validate_oauth:validate(Body),
+            [
+                %% Current behavior: split_path splits the dotted key, lookup fails,
+                %% no scopes extracted -> authz_unverified. This will flip to ok once
+                %% upstream fixes the lookup.
+                ?_assertMatch({error, authz_unverified, _}, Result),
+                ?_assert(reason_contains(Result, "no scopes for this resource_server"))
```
lukebakken: These two assertions are byte-identical to `authz_no_effective_scopes_after_prefix_test_` (line 994-995), which pins the same `{error, authz_unverified, _}` + "no scopes for this resource_server" outcome for a completely different cause (a token whose scopes all have the wrong prefix).

Because any `additional_scopes_key` that fails to resolve for any reason (key renamed, feature removed, value silently ignored) yields the same category and message, this pin cannot distinguish the dotted-key `split_path` mis-split from unrelated no-scopes outcomes. It only detects a change once effective scopes become non-empty, not when the dotted-key path breaks in some other way. If the goal is to isolate the mis-split specifically, consider asserting on something that is unique to this path (for example, that a control token carrying the same scopes under a non-dotted key does reach `ok`, so the dotted vs non-dotted key is the only difference).

test/aws_auth_validate_oauth_tests.erl
```diff
@@ -1032,6 +1032,66 @@ authz_iam_scope_alias_grants_access_test_() ->
...
+        maybe_skip_authz(fun() ->
+            #{jwk_pub := PubJwk, sign := Sign} = rsa_signer(<<"k1">>),
+            %% Token carries scopes under a dotted claim key (OIDC/STS-style URI).
+            %% split_path/1 will incorrectly split this on dots, breaking resolution.
+            Token = Sign(#{
+                <<"exp">> => future(),
+                <<"aud">> => <<"rabbitmq">>,
+                <<"https://sts.amazonaws.com/tags">> => [
+                    <<"rabbitmq.write:*/*">>, <<"rabbitmq.read:*/*">>
+                ]
```
lukebakken: The flip-to-`ok` contract the comment describes (lines 1049-1052) depends on the exact shape of the eventual upstream fix, and a plausible alternative fix would leave this pin red after #16947 is resolved.

The value here is already `rabbitmq.`-prefixed. It flips to `ok` only if the fix returns this list verbatim before the `scope_prefix` "rabbitmq." strip. If upstream instead resolves `additional_scopes_key` values through the `resource_server_id` map-prefixing branch (prepending "rabbitmq."), the scopes become `rabbitmq.rabbitmq.write:*/*`, the strip leaves `rabbitmq.write:*/*`, `resource_access` is `false`, and the pin stays `authz_unverified` even though the drift was fixed. Using an unprefixed value here (e.g. `write:*/*`, letting the prefix logic add `rabbitmq.`) would make the flip robust to either fix shape.

lukebakken (commented):
Reviewed 7e07cde. The doc-comment rewrite is accurate (`authz_check` and the four authz fields are real entries in `allowed_fields/0`, the "no scopes for this resource_server" reason string exists in `aws_auth_validate_oauth_authz.erl`, and the `split_path/1` characterization matches upstream verbatim), and the new pin's mechanism is real where it runs.

The comments below are all about the pin test's efficacy as a drift detector, not a logic bug in shipped code (the change is test-only plus a comment). Three ways the pin can miss the drift it exists to catch. None are blocking; whether they matter is a judgment call for a deliberate parity pin, but they are worth weighing before merge.

test/aws_auth_validate_oauth_tests.erl
sdewhitt: I have an extra control token dotted vs. non-dotted key test parked on #161

test/aws_auth_validate_oauth_tests.erl
sdewhitt: Verified against the arity-4 backend (`4.2.0-beta.4.869`). `extract_scope_list_from_token_value/2` branches on the claim value's *type*, not the fix shape: a **list** value is taken verbatim, only a **map** keyed by `resource_server_id` gets prefixed. Running `normalize_token_scope` with `scope_prefix = "rabbitmq."`:

| claim value | normalized (post-strip) | result |
|---|---|---|
| `["rabbitmq.write:*/*"]` (current) | `["write:*/*"]` | flips to `ok` ✅ |
| `["write:*/*"]` (unprefixed) | `[]` | stays `authz_unverified` ❌ |
| `#{"rabbitmq" => ["write:*/*"]}` | `["write:*/*"]` | flips to `ok` ✅ |

So the unprefixed value would leave the pin permanently red — with a list value there's no prefixing branch to re-add `rabbitmq.`, so the strip filters it to `[]`. The double-prefix case only arises for a map value, which this token doesn't use. Keeping the prefixed value (it flips correctly) and adding a comment that list-typed values resolve verbatim. Caveat: verified against the current dep; a re-check is implied by the parity bump if merged #16985 changes this fn.

test/aws_auth_validate_oauth_tests.erl
sdewhitt: Agreed. On a checkout without the arity-4 API, `HAVE_OAUTH2_RESOURCE_SERVER` is undefined, `available/0` is `false`, and `maybe_skip_authz/1` returns `[]`, so the body never runs — the comment's "the moment the dependency changes" overstates it. Your backport scenario (split_path fix on a v4.3.x without the arity-4 scope API → pin stays dormant → regression unnoticed) is real. I'll amend the comment to state the pin is dormant until the arity-4 scope API is present, so green isn't misread as "still broken as expected."

the-mikedavis (approved):

test/aws_auth_validate_oauth_tests.erl
lukebakken: Resolved by 6d0df71. The DORMANCY paragraph states the precondition explicitly (the pin only runs once the dep ships resource_access/4), spells out the exact skip mechanism (HAVE_OAUTH2_RESOURCE_SERVER undefined -> available/0 false -> maybe_skip_authz/1 returns []), and names the consequence I was worried about: a green suite does not mean "still broken as expected." It also keeps the split_path-backport-without-arity-4 case and points at the parity bump as the real backstop. That is the right call for a doc-level fix. Marking resolved.

test/aws_auth_validate_oauth_tests.erl
lukebakken: Resolved by 6d0df71, and the FLIP SHAPE paragraph correctly refutes my suggestion. I verified against the resolved dep: rabbit_auth_backend_oauth2:extract_scope_list_from_token_value/2 returns a LIST value verbatim (line 344), and only the map clause keyed by resource_server_id prefix-injects (lines 347-354). So the double-prefix scenario I described cannot arise from a list value, and my proposed unprefixed value (write:*/*) would in fact be filtered to [] by the scope_prefix strip and never flip. Keeping the prefixed value is correct; the comment now records why. My finding was wrong on this one. Marking resolved.

test/aws_auth_validate_oauth_tests.erl
lukebakken: Still open in 6d0df71: the two assertions are unchanged and no control case was added, so the pin still cannot distinguish the dotted-key mis-split from any other no-scopes outcome. It shares its exact assertion pair with authz_no_effective_scopes_after_prefix_test_.

Not blocking, and I understand the tradeoff differently now that the DORMANCY note names the parity bump as the real backstop: if the revisit is what actually catches the drift, the assertion precision matters less. Two options if you want to close the gap cheaply, otherwise feel free to resolve as accepted-as-is:

- Add a control assertion in the same generator: the same scopes under a NON-dotted key (e.g. `tags`) reach `ok`, making the dotted vs non-dotted key the only variable.
- Or note in the comment that the assertion pair is intentionally shared and that the flip to `ok`, not the current message, is the real signal.

lukebakken (commented):
Re-reviewed at 6d0df71 (the pin commit is now 79b7c9f after the amend). PR scope is unchanged: `src/aws_auth_validate_oauth.erl` plus `test/aws_auth_validate_oauth_tests.erl`, no production-code behavior change.

## Round-1 findings

- **Dormancy (resolved).** The new DORMANCY paragraph names the precondition, the exact skip mechanism, and the "green does not mean still-broken-as-expected" consequence, and points at the parity bump as the real backstop.
- **Flip shape (resolved, my finding was wrong).** The FLIP SHAPE paragraph is correct and I verified it against the resolved dep: `extract_scope_list_from_token_value/2` returns a list value verbatim, and only the `resource_server_id`-keyed map clause prefix-injects. My suggested unprefixed value would have been stripped to `[]` and never flipped. Keeping the prefixed value is right.
- **Shared assertion pair (still open).** Unchanged, and no control case was added. Left open with two cheap options in the thread; not blocking, and less important now that the comment names the parity bump as the actual backstop.

## Checks

erlfmt clean on both files at this head. The two `actions/checkout` v7.0.1 bumps visible when diffing against the old commit came in from main, not from this PR.

No further findings from the update. The added comment blocks are accurate against the dep as resolved on this branch.

test/aws_auth_validate_oauth_tests.erl
lukebakken: Addressed in 80ea911. The body is now a `Validate(ClaimKey)` closure so the claim key is the only variable between the two cases:

- dotted key -> `authz_unverified` + the same reason substring (the pin, unchanged)
- non-dotted key (`tags`) -> `ok` (the new control)

That is what makes the pin diagnostic: if the control ever stops granting, the breakage is in `additional_scopes_key` resolution generally rather than in the dotted-key split, which the shared assertion pair alone could not distinguish.

Since the pin is dormant on both local checkouts (no `resource_access/4` in the resolved dep), a passing suite would not have exercised the new assertion, so I verified it by driving the real upstream path directly:

```
DOTTED  -> {effective_scopes,[],                     granted,false}
CONTROL -> {effective_scopes,[read:*/*, write:*/*],  granted,true}
```

`normalize_token_scope/2` -> `resource_access/3` on the resolved dep, so the control genuinely grants. Also added a CONTROL paragraph to the header comment. erlfmt clean, 98 eunit tests pass. Marking resolved.

lukebakken (approved):
Approving at 80ea911. All three round-1 threads are resolved and CI is green across all four broker series (build-test main / v4.3.x / v4.2.x / v3.13.7, plus format-check).

Where the three findings landed:

- **Dormancy**: documented. The DORMANCY paragraph names the precondition, the skip mechanism, and the "green does not mean still-broken-as-expected" consequence, and points at the parity bump as the real backstop.
- **Flip shape**: documented, and the note correctly refuted my suggestion. `extract_scope_list_from_token_value/2` returns a list value verbatim; only the `resource_server_id`-keyed map clause prefix-injects, so the prefixed value is the right choice and an unprefixed one would never flip.
- **Shared assertion pair**: fixed in 80ea911 with a paired non-dotted control case, so the claim key is the only variable and the pin is now diagnostic rather than sharing its signature with the wrong-prefix test.

One caveat worth recording for whoever revisits this at the parity bump: the pin is dormant in CI too, not just locally. Test counts are identical before and after 80ea911 (718 on main, 725 on v4.3.x), and since that commit adds exactly one assertion, an executing generator would have incremented them. No series in CI currently ships the arity-4 scope API, so the control assertion has never run in CI. It was verified instead by driving the resolved dependency directly:

```
DOTTED  -> {effective_scopes,[],                     granted,false}
CONTROL -> {effective_scopes,[read:*/*, write:*/*],  granted,true}
```

That is `normalize_token_scope/2` into `resource_access/3`, so the control genuinely grants and the dotted key genuinely does not. The practical consequence is that CI cannot regress-test this pin until the dependency ships `resource_access/4`; the parity bump remains the moment to re-run and confirm the flip.

No production code changes in this PR, so no runtime risk.

PR #165, opened 2026-07-27
TLS Validation Enhancements

sdewhitt:
The `tls` method validated CA material only (ARN resolve, PEM well-formedness, expiry, option shapes), so every failure mode of the publicly documented SSL certificate authentication setup (`auth_mechanisms.1 = EXTERNAL` + `ssl_cert_login_from`, per the Amazon MQ SSL/mTLS tutorials) stayed invisible until after a broker restart: a client cert that does not chain to the configured CA bundle, a `cert_login` mode that extracts no name from the cert, `verify_none` silently disabling EXTERNAL entirely, and a CN with no matching internal user -- the SSL tutorial's own top documented footgun (its prerequisite note requires the cert CN to match the username). The TLS handshake succeeds in most of these, so operators debug the wrong layer.

Following the oauth `access_token` precedent (caller-minted, non-secret proof material activating optional validation layers), a request may now carry a `client_cert` PEM and a `cert_login` block mirroring `ssl_cert_login_from` / `ssl_cert_login_san_type` / `ssl_cert_login_san_index`. Both are optional; a request without them behaves exactly as before.

### 1. Layer 1 -- chain validation (`client_cert`)
- `public_key:pkix_path_validation/3` of the supplied leaf (+ optional intermediates) against the resolved `cacertfile_arn` bundle, honoring an explicitly supplied `depth`. This answers "will the broker trust certs issued the way mine are?" against what the ARN *actually resolves to* -- which local `openssl verify` cannot.
- Not chaining / expired chain member -> `auth_failed` (422); malformed -> `input_invalid`. Fixed reasons; no issuer/subject/serial ever echoed (R4).
- **R6 guard:** the PEM must contain only CERTIFICATE entries. Any private-key entry (`RSAPrivateKey`, `DSAPrivateKey`, `ECPrivateKey`, `PrivateKeyInfo`, `EncryptedPrivateKeyInfo`) is rejected with a fixed reason telling the caller never to send key material. Chain validation and name extraction need public material only; possession-proof is deliberately out of scope.

### 2. Layer 2 -- username-extraction parity (`cert_login`)
- Mirrors the three-clause dispatch of `rabbit_ssl:peer_cert_auth_name/2` (`distinguished_name` / `common_name` incl. the multi-CN `","` join / `subject_alternative_name` incl. the `subject_alt_name` alias, the `otp_san_type/1` mapping, 0-based `san_index`, and `otherName` sanitization) while delegating the actual extraction to the broker's pure `rabbit_cert_info` helpers, so the extraction logic itself cannot drift. A coupling note documents the mirrored dispatch.
- `peer_cert_auth_name/2` is deliberately NOT called: its `auth_config_sane/0` reads the LIVE broker's `ssl_options.verify` app env -- the wrong input for validating a candidate config. Instead, `cert_login` requires the *supplied* `ssl_options.verify = verify_peer` (`config_conflict` otherwise), reaching the same verdict the live broker's unsafe-config refusal would.
- Extraction yielding no name (mode/cert mismatch, SAN index past the end, no CN) -> `auth_failed` with a fixed stage-naming reason. This is the check nothing customer-side can replicate: the extraction logic is RabbitMQ-internal and the only oracle today is a restarted broker.

### 3. Layer 3 -- user resolution
- Read-only `rabbit_auth_backend_internal:exists/1` on the raw extracted name (not `check_user_login/2`, which runs the full backend chain and can touch caches). Guarded by an `ensure_loaded`/`function_exported` probe (the `aws_auth_validate_oauth_authz` pattern) -> `config_conflict` if the API is absent on a broker series.
- Not found -> `auth_failed`; the reason echoes the extracted username (length-capped, control chars stripped). R4 basis: the name derives solely from the caller's own supplied certificate, the same footing as the oauth token reasons. The *lookup* uses the raw name so the verdict cannot diverge from what the live broker's EXTERNAL login would do; sanitization applies only to the echo.
- R3 note: this is the endpoint's first read of live broker state (the internal user table) -- read-only, no mutation, and a documented, deliberate deviation from pure-config validation.

### Cross-field rules and ordering
- `cert_login` requires `client_cert`; `client_cert` alone is allowed (chain-check only -- the mTLS-encryption-only tutorial case).
- All new input checks (PEM decode, `cert_login` shape, cross-field conflicts) run in the pure phase, preserving ARN-first ordering: a malformed request never triggers an AssumeRole or secret fetch.
- No new network surface: both tutorials' configs are inbound-listener material, so validation is pure computation plus the existing ARN fetch. No SSRF surface; DER parsing is bounded by the existing 64 KiB body cap and the concurrency semaphore.

### Testing
- `erlfmt -c` clean on all `.erl` changes.
- Full umbrella build passes; full clean eunit run green (773 tests, including meck'd Layer-3 found / not-found / API-absent cases and generated self-signed CA -> leaf chain and SAN/CN extraction cases).
- `ct-aws_auth_validate_tls`: 11 passed, 0 failed (adds end-to-end 204 valid-chain, 422 bad-chain, and 422 cert-login-conflict cases through the registry).
- Backward compat: requests without the new fields exercise the unchanged path; all pre-existing tls tests pass unmodified.

### DIFF
--- a/src/aws_auth_validate_tls.erl
+++ b/src/aws_auth_validate_tls.erl
@@ -501,6 +501,20 @@ find_non_cert_entry([{Type, _Der, _Enc} | _Rest]) ->
 
 %% Returns true when the PEM type tag identifies a private key -- either an
 %% atom from the traditional set or the tuple tag used by OpenSSH keys.
+%%
+%% The {no_asn1, _} clause is reachable at RUNTIME even though dialyzer claims
+%% otherwise: public_key's exported pem_entry() type enumerates only the ASN.1
+%% record tags, but pubkey_pem emits {no_asn1, new_openssh} for an OpenSSH
+%% private key block. Verified on this OTP:
+%%
+%%   public_key:pem_decode(<<"-----BEGIN OPENSSH PRIVATE KEY-----"...>>)
+%%   => [{{no_asn1, new_openssh}, <<...>>, not_encrypted}]
+%%
+%% So the incomplete upstream spec, not this clause, is the inaccuracy. The
+%% clause is what keeps an OpenSSH key from being reported as a generic
+%% unexpected entry instead of key material, so it must not be removed. The
+%% nowarn is scoped to this function only.
+-dialyzer({nowarn_function, is_known_key_type/1}).
 is_known_key_type(Type) when is_atom(Type) ->
     lists:member(Type, ?PRIVATE_KEY_TYPES);
 is_known_key_type({no_asn1, _}) ->
--- a/src/aws_auth_validate_tls.erl
+++ b/src/aws_auth_validate_tls.erl
@@ -12,23 +12,32 @@
 %%
 %% It validates the material only, not a live handshake. The other backends
 %% probe an outbound auth server; here the config is an inbound listener, so
-%% there is nothing for the broker to connect to (and no client keystore to
-%% connect with). Checks:
+%% there is nothing for the broker to connect to. Checks:
 %%   1. the cacertfile ARN resolves,
 %%   2. the resolved PEM holds at least one well-formed CA certificate,
 %%   3. none of those certificates is expired or not yet valid,
 %%   4. verify / fail_if_no_peer_cert / depth / versions are well-shaped.
 %%
-%% A pass means the material is usable, not that mTLS as a whole works: it does
-%% not cover client-cert chaining, common-name-to-user mapping, network
-%% reachability, or whether the broker is actually running this config.
+%% Optionally, when `client_cert' and `cert_login' are supplied, the backend
+%% also validates:
+%%   5. the client certificate chains to the supplied CA bundle (Layer 1),
+%%   6. a username can be extracted from the leaf cert using the configured
+%%      cert_login strategy -- DN / CN / SAN (Layer 2),
+%%   7. the extracted username resolves to an internal broker user (Layer 3).
+%%
+%% A pass means the material is usable and (when layers 5-7 are exercised) the
+%% cert-login pipeline produces a resolvable user, not that the broker is
+%% actually running this config.
 %%
 %% Result categories (shared with the other backends):
 %%   * input_invalid (400) -- bad target/ssl_options, missing cacertfile_arn,
 %%     ARN resolve failure, or a PEM with no parseable CA certificate.
 %%   * tls_failed (400) -- a CA certificate is expired or not yet valid.
+%%   * auth_failed (422) -- client certificate chain validation failed, no
+%%     username could be extracted, or the extracted user does not exist.
 %%   * config_conflict (422) -- a cacertfile_arn is given but no
-%%     aws.arns.assume_role_arn is configured.
+%%     aws.arns.assume_role_arn is configured; or cert_login references
+%%     unavailable broker modules.
 -module(aws_auth_validate_tls).
 
 -behaviour(aws_auth_validate_backend).
@@ -36,14 +45,18 @@
 -export([method_name/0, validate/1, allowed_fields/0]).
 
 -ifdef(TEST).
-%% Exposed for the unit tests: the pure input parser and the certificate-validity
+%% Exposed for the unit tests: the pure input parser, the certificate-validity
 %% helpers (classify_validity/3 lets the expired/not-yet-valid branches be tested
-%% without generating an actually-expired cert).
+%% without generating an actually-expired cert), and the client-cert-login layers.
 -export([
     parse_input/1,
     check_cert_validity/1,
     cert_validity_seconds/1,
-    classify_validity/3
+    classify_validity/3,
+    validate_chain/3,
+    extract_username/2,
+    resolve_user/1,
+    sanitize_username/1
 ]).
 -endif.
 
@@ -94,6 +107,71 @@
     "set aws.arns.assume_role_arn"
 >>).
 
+%% client_cert / cert_login reason strings (Layer 1-3 validation).
+-define(REASON_BAD_CLIENT_CERT,
+    <<"client_cert must be a non-empty PEM-encoded certificate chain">>
+).
+-define(REASON_CLIENT_CERT_PRIVATE_KEY,
+    <<"client_cert must not contain private key material; send only certificate entries">>
+).
+-define(REASON_BAD_CERT_LOGIN, <<"cert_login must be an object">>).
+-define(REASON_BAD_CERT_LOGIN_FROM,
+    <<"cert_login.from must be distinguished_name, common_name, or subject_alternative_name">>
+).
+-define(REASON_SAN_TYPE_REQUIRES_SAN_FROM,
+    <<"cert_login.san_type is only valid when from is subject_alternative_name">>
+).
+-define(REASON_SAN_INDEX_REQUIRES_SAN_FROM,
+    <<"cert_login.san_index is only valid when from is subject_alternative_name">>
+).
+-define(REASON_BAD_SAN_TYPE,
+    <<"cert_login.san_type must be dns, ip, email, uri, or other_name">>
+).
+-define(REASON_BAD_SAN_INDEX,
+    <<"cert_login.san_index must be a non-negative integer">>
+).
+-define(REASON_UNKNOWN_CERT_LOGIN_KEY,
+    <<"cert_login contains an unknown key; allowed keys are from, san_type, san_index">>
+).
+-define(REASON_CERT_LOGIN_REQUIRES_CLIENT_CERT,
+    <<"cert_login requires client_cert to be present">>
+).
+-define(REASON_CERT_LOGIN_REQUIRES_VERIFY_PEER,
+    <<"cert-based login requires ssl_options.verify = verify_peer">>
+).
+-define(REASON_CHAIN_FAILED,
+    <<"the client certificate does not chain to the supplied CA bundle">>
+).
+-define(REASON_CHAIN_EXPIRED,
+    <<"the client certificate chain contains an expired or not-yet-valid certificate">>
+).
+-define(REASON_CHAIN_MALFORMED,
+    <<"a certificate in the client chain could not be parsed">>
+).
+-define(REASON_EXTRACT_FAILED,
+    <<"no username could be extracted from the client certificate with the supplied cert_login settings">>
+).
+-define(REASON_USER_NOT_FOUND(Name),
+    iolist_to_binary([
+        <<"no internal user named ">>, sanitize_username(Name), <<" exists">>
+    ])
+).
+-define(REASON_USER_LOOKUP_UNAVAILABLE,
+    <<"user-resolution check unavailable on this broker series">>
+).
+
+-define(INTERNAL_BACKEND, rabbit_auth_backend_internal).
+-define(MAX_USERNAME_LEN, 256).
+
+%% Private key entry types that must be rejected from client_cert input (R6).
+-define(PRIVATE_KEY_TYPES, [
+    'RSAPrivateKey',
+    'DSAPrivateKey',
+    'ECPrivateKey',
+    'PrivateKeyInfo',
+    'EncryptedPrivateKeyInfo'
+]).
+
 %% Surface passed to the shared aws_auth_validate_ssl helpers: the ARN-bearing
 %% keys, the allowed-key set, and this backend's reason strings. client_cert is
 %% false (no client pair) and sni_key is unused here; both are required by the
@@ -125,7 +203,7 @@ method_name() ->
     <<"tls">>.
 
 allowed_fields() ->
-    [<<"target">>, <<"ssl_options">>].
+    [<<"target">>, <<"ssl_options">>, <<"client_cert">>, <<"cert_login">>].
 
 -spec validate(map()) -> aws_auth_validate_backend:result().
 validate(Body) when is_map(Body) ->
@@ -149,7 +227,10 @@ parse_input(Body) ->
     Steps = [
         fun parse_target/2,
         fun parse_ssl_options/2,
-        fun require_cacert/2
+        fun require_cacert/2,
+        fun parse_client_cert/2,
+        fun parse_cert_login/2,
+        fun cross_field_checks/2
     ],
     run_steps(Steps, Body, #{}).
 
@@ -193,7 +274,9 @@ require_cacert(_Body, #{ssl_options := Map} = Acc) ->
 %%--------------------------------------------------------------------
 
 %% Resolve the cacertfile ARN, then decode and check the CA bundle. The only
-%% network call is the ARN fetch; nothing connects to a listener.
+%% network call is the ARN fetch; nothing connects to a listener. When client
+%% certificate layers are present, continue into chain/extract/resolve after
+%% the CA bundle passes.
 do_tls_validate(#{ssl_options := Map} = Params) ->
     %% A request that referenced no ARN carries the `none' sentinel, which
     %% resolve_arn/2 refuses. cacertfile_arn is required, so a valid request
@@ -204,7 +287,10 @@ do_tls_validate(#{ssl_options := Map} = Params) ->
         {error, _} ->
             {error, input_invalid, ?REASON_ARN_RESOLVE};
         {ok, Pem} ->
-            decode_and_check(Pem)
+            case decode_and_check(Pem) of
+                {error, _, _} = Err -> Err;
+                ok -> maybe_chain_validate(Params, Pem)
+            end
     end.
 
 %% Decode the CA PEM and check each certificate. The decode is wrapped because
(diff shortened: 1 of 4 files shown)


the-mikedavis (approved):

src/aws_auth_validate_tls.erl
```diff
@@ -306,3 +392,356 @@ ymd_to_seconds(Year, Rest) ->
...
+                undefined -> [];
+                N when is_integer(N), N >= 0 -> [{max_path_length, N}]
+            end,
+        %% pkix_path_validation wants: TrustAnchor (DER), Chain (Options).
+        %% ClientDers is leaf-first ([leaf, intermediates...]); the topmost cert
+        %% (lists:last) is the one issued by a bundle CA, so match the anchor
+        %% against its issuer. The chain itself must be passed anchor-closest
+        %% first (the cert the anchor issued down to the leaf), which is the
+        %% reverse of the leaf-first ClientDers.
+        TopCert = lists:last(ClientDers),
```
lukebakken: **False FAIL (4 of 7 divergences): a leaf-first but unordered chain is rejected where the broker accepts it.**

`lists:last(ClientDers)` assumes the chain is not just leaf-first but fully ordered. A PEM that is leaf-first with an arbitrarily ordered tail, for example `[leaf, root, int]`, makes `lists:last` return the intermediate, so the anchor is matched against the wrong issuer.

Measured against a live listener with `cacerts = [Root]`, that exact certfile completes the handshake, because `ssl_certificate:paths/2` and `unorded_or_extraneous/2` rebuild the tail from the peer cert. This endpoint returns 422 `auth_failed` "the client certificate does not chain to the supplied CA bundle" for the same bytes.

One narrowing worth recording, since I checked it specifically: a fully root-first paste is rejected by the broker too (`{alert, bad_certificate}`), so that case is a shared rejection and not a false failure. `ssl_certificate.erl`'s own header comment says the peer cert must still be first. So the divergence is only over tail ordering.

Separately, the leaf-first requirement is invisible to the caller. It appears only in implementation comments here and on lines 577 and 592. `?REASON_CLIENT_CERT_INVALID` says only that `client_cert` must be a non-empty PEM-encoded certificate chain, and no doc in the plugin mentions ordering. Even if you decide tail reordering is out of scope, the field's error string is the place to state the requirement.

src/aws_auth_validate_tls.erl
```diff
@@ -306,3 +392,356 @@ ymd_to_seconds(Year, Rest) ->
...
+        Chain = lists:reverse(ClientDers),
+        case find_trust_anchor(TopCert, CaDers) of
+            {ok, AnchorDer} ->
+                case public_key:pkix_path_validation(AnchorDer, Chain, Opts) of
+                    {ok, _} ->
+                        ok;
+                    {error, {bad_cert, cert_expired}} ->
+                        {error, auth_failed, ?REASON_CHAIN_EXPIRED};
+                    {error, {bad_cert, {cert_expired, _}}} ->
```
lukebakken: **Cleanup: this clause is dead, and dialyzer can prove it.**

```erlang
{error, {bad_cert, {cert_expired, _}}} ->
```

Dialyzer reports that `{'error',{'bad_cert',{'cert_expired',_}}}` can never match the return type of `pkix_path_validation/3`. In the OTP 27 tree this checkout resolves, the sole emission site (`pubkey_cert.erl:161`) uses the bare atom, `bad_cert_reason()` in `public_key.erl:309` declares `cert_expired` as an atom, and no tuple-wrapped form exists anywhere in the OTP sources.

The one route that could produce an arbitrary tuple reason is a caller-supplied `verify_fun` returning `{fail, UserReason}`, and `validate_chain/3` passes none (`Opts` is only `[]` or `[{max_path_length, N}]`), so the default verify fun applies and the tuple is unreachable here.

Deleting lines 606 and 607 loses nothing: the `{error, {bad_cert, _}}` catch-all below already absorbs any unforeseen shape. Worth doing mainly so a future reader does not add more tuple variants on the same false premise, and so `gmake dialyze` over this component stays quiet.

src/aws_auth_validate_tls.erl
```diff
@@ -306,3 +392,356 @@ ymd_to_seconds(Year, Rest) ->
...
+                        {error, auth_failed, ?REASON_CHAIN_EXPIRED};
+                    {error, {bad_cert, _}} ->
```
lukebakken: **False diagnosis (6 of 8): `max_path_length_reached` is reported as "does not chain", and the new depth test cannot catch it.**

This catch-all collapses every remaining `bad_cert` reason into `?REASON_CHAIN_FAILED`, "the client certificate does not chain to the supplied CA bundle". For a depth violation that statement is actively false: the certificate does chain, the operator just capped the path length.

Verified empirically with an openssl root/intermediate/leaf set:

```
depth 0, leaf+int chain:  {error,{bad_cert,max_path_length_reached}}
no depth, same chain:     {ok, ...}
depth 1, same chain:      {ok, ...}
```

Same chain, same bundle, so it demonstrably chains. The operator gets sent hunting a nonexistent CA-bundle problem instead of raising `depth`. The pass/fail verdict is right (a broker with `depth = 0` would also reject), so this is a diagnostic-accuracy defect rather than a wrong answer, which matters more than usual in an endpoint whose entire purpose is diagnosis. The same clause also swallows `invalid_key_usage`, `missing_basic_constraint` and `invalid_issuer` into the identical string, so a dedicated `max_path_length_reached` clause fixes only the one instance; a reason that names the underlying atom would cover all of them.

The test added in 9970e91 is the one loosened assertion in the PR, so nothing pins the wrong message:

```erlang
tls_validate_chain_intermediate_depth_zero_rejected_test() ->
    ?assertMatch(
        {error, auth_failed, _},
```

All three sibling chain tests pin the reason prefix, for example `{error, auth_failed, <<"the client certificate does not chain", _/binary>>}`. Pinning this one too would have surfaced the mismatch.

src/aws_auth_validate_tls.erl
```diff
@@ -306,3 +392,356 @@ ymd_to_seconds(Year, Rest) ->
...
+                end;
+            not_found ->
+                {error, auth_failed, ?REASON_CHAIN_FAILED}
+        end
+    catch
+        _:_ -> {error, input_invalid, ?REASON_CHAIN_MALFORMED}
+    end.
+
+%% Find the trust anchor in CaDers whose subject matches the issuer of TopCertDer.
+find_trust_anchor(TopCertDer, CaDers) ->
```
lukebakken: **Highest severity: this is a false PASS, and it greenlights the exact misconfiguration the endpoint exists to catch.**

`find_trust_anchor/2` matches a bundle entry by normalized issuer/subject name and nothing else. There is no `pkix_is_self_signed` check anywhere in the module, and `decode_and_check/1` never inspects `basicConstraints`, so the matched bundle entry need not be self-signed or even be a CA.

`public_key:pkix_path_validation/3` will happily treat a non-self-issued cert as an anchor, so an intermediate-only `cacertfile` bundle returns `ok` here and the endpoint answers HTTP 204 "material is usable".

A real listener with that same bundle rejects every client. I stood up `ssl:listen` with `{cacerts, [IntDer]}`, `verify_peer`, `fail_if_no_peer_cert`, and connected a real client:

```
bundle=[Int], client sends [leaf]        -> {alert, unknown_ca}
bundle=[Int], client sends [leaf, int]   -> {alert, unknown_ca}
bundle=[Root], client sends [leaf, int]  -> handshake ok   (control)
bundle=[Root, Int], client sends [leaf]  -> handshake ok   (control)
```

`ssl_certificate:handle_partial_chain/4` only accepts an anchor once `public_key:pkix_is_self_signed/1` returns true; the escape hatch is a `partial_chain` fun, and the only one in the tree belongs to `rabbitmq_trust_store`, so it is not in play for a plain `ssl_options` listener.

An operator who forgets the root in their bundle is told their material is fine and then finds no client can connect, which is worse than no validation at all. Requiring the matched anchor to be self-signed (absent a `partial_chain` fun) would align the verdict with the handshake.

src/aws_auth_validate_tls.erl
```diff
@@ -306,3 +392,356 @@ ymd_to_seconds(Year, Rest) ->
...
+    TopOtp = public_key:pkix_decode_cert(TopCertDer, otp),
+    Issuer = TopOtp#'OTPCertificate'.tbsCertificate#'OTPTBSCertificate'.issuer,
+    find_by_subject(Issuer, CaDers).
+
+find_by_subject(_Issuer, []) ->
+    not_found;
+find_by_subject(Issuer, [CaDer | Rest]) ->
```
lukebakken: **False FAIL: a CA key rollover bundle succeeds or fails purely on PEM order.**

`find_by_subject/2` returns the *first* subject-DN match and `validate_chain/3` never retries with the remaining candidates. Two CAs sharing a subject DN but holding different keys is not an edge case, it is what a root key rollover looks like, and having both in the bundle is the whole reason they are both there.

Reproduced with real certs, both roots `CN=TestRoot`, leaf chaining to the older one:

```
bundle=[Old, New] -> {bad_cert, invalid_signature} -> 422 "does not chain to the supplied CA bundle"
bundle=[New, Old] -> ok
```

The live broker does not behave this way: `ssl_handshake:path_validate/9` iterates every candidate path from `ssl_certificate:trusted_cert_and_paths/4`, and `extraneous_chains/1` will try up to 16, only returning the accumulated error once all candidates fail. I confirmed a listener with the byte-identical bundle and client chain completes the handshake in either order.

So the verdict flips on the order the operator happened to concatenate two PEMs, and the message blames chaining when the real result was a signature mismatch against the wrong twin. Collecting all name matches and accepting if any validates would match the broker.

src/aws_auth_validate_tls.erl
```diff
@@ -306,3 +392,356 @@ ymd_to_seconds(Year, Rest) ->
...
+extract_username(LeafDer, #{from := subject_alternative_name, san_type := Type, san_index := Index}) ->
+    OtpType = otp_san_type(Type),
+    SANs = rabbit_cert_info:subject_alternative_names(LeafDer),
+    Filtered = [V || {T, V} <- SANs, T =:= OtpType],
+    %% Index is 0-based in config, 1-based for lists:nth.
+    case length(Filtered) > Index of
+        true ->
+            Raw = lists:nth(Index + 1, Filtered),
+            Value = maybe_sanitize_other_name(OtpType, Raw),
+            {ok, iolist_to_binary([Value])};
```
lukebakken: **Uncaught `badarg`: a caller-supplied cert turns a 422 into a bare HTTP 500.**

```erlang
{ok, iolist_to_binary([Value])};
```

`rabbit_cert_info:sanitize_other_name/1` *returns* its failure rather than raising. Its case falls through to an `Other -> Other` branch, so a value that is not a decodable `DirectoryString` comes back as an error tuple. I ran it directly:

```erlang
> rabbit_cert_info:sanitize_other_name(<<22,11,"foo@bar.com">>).
{error,{asn1,{{invalid_choice_tag,{22,<<"foo@bar.com">>}}, ...}}}
> iolist_to_binary([{error, {asn1, ...}}]).
** exception error: bad argument
```

That is reachable with `cert_login = {from: subject_alternative_name, san_type: other_name}` and an `otherName` holding a non-DirectoryString value, for example `openssl ... otherName=1.2.3.4;IA5STRING:foo@bar.com`. UPN and hardware-identity `otherName`s are commonly encoded this way.

Nothing on the path catches it. `do_tls_validate` -> `maybe_chain_validate` -> `maybe_extract_username` -> `extract_username` has no `try`, and `validate_chain/3`'s `try` does not enclose this call. So the `badarg` escapes to `aws_auth_validate_mgmt:with_semaphore/7`'s catch-all and the caller gets HTTP 500 `internal_error` with no reason at all, rather than a 422 `auth_failed` saying the username could not be extracted. Every other extraction failure in this module is a categorized result; this one is not.

Worth noting the comparison the code cites: upstream `rabbit_ssl:peer_cert_auth_name/2` returns the same error tuple onward and never passes it to `iolist_to_binary`, so the crash is introduced here rather than inherited. Matching on the tuple and returning `?REASON_EXTRACT_FAILED` would restore the contract.

src/aws_auth_validate_tls.erl
```diff
@@ -306,3 +392,356 @@ ymd_to_seconds(Year, Rest) ->
...
+resolve_user(Username) ->
+    case internal_backend_available() of
+        false ->
+            {error, config_conflict, ?REASON_USER_LOOKUP_UNAVAILABLE};
+        true ->
+            %% Look up the RAW extracted name -- it is what the broker's own
+            %% EXTERNAL login would pass to check_user_login. Sanitization is
+            %% applied only when echoing the name in the reason (R4), so the
+            %% lookup verdict cannot diverge from the live broker's.
+            case ?INTERNAL_BACKEND:exists(Username) of
```
lukebakken, line 977: **False FAIL, and the comment just above it is not accurate.**

The comment on lines 717 to 720 says the lookup verdict "cannot diverge from the live broker's". It can, because the broker's EXTERNAL login does not go through `rabbit_auth_backend_internal` specifically. `rabbit_auth_mechanism_ssl.erl:71` calls `rabbit_access_control:check_user_login(Username, [])`, which reads the `auth_backends` env and `lists:foldl`s over every configured module (`rabbit_access_control.erl:177-187`).

So on a broker with `auth_backends.1 = ldap` (or oauth2, or http) and users living only in that backend, an mTLS login for `CN=svc-app` succeeds live while this endpoint returns 422 `auth_failed` "no internal user named svc-app exists". The operator is told a working cert-login config is broken.

What makes this one harder to notice than the other divergences: it fails silently by omission rather than by crashing. `internal_backend_available/0` returns true whenever the module is loadable and exports `exists/1`, which it always does, so the `config_conflict` escape hatch on line 715 never fires and nothing in the response hints that only one of several configured backends was consulted.

The sanitization half of the comment is correct and I am not questioning it (see the separate note on `sanitize_username/1`). It is specifically the non-divergence claim that needs either a fix or a rewrite, because as written it will discourage the next reader from questioning the single-backend lookup.

src/aws_auth_validate_tls.erl
```diff
@@ -306,3 +392,356 @@ ymd_to_seconds(Year, Rest) ->
...
+%% the caller's own supplied certificate (R4 basis for echoing it), but we still
+%% bound it against degenerate inputs.
+-spec sanitize_username(binary()) -> binary().
+sanitize_username(Name) when is_binary(Name) ->
+    Capped =
+        case byte_size(Name) > ?MAX_USERNAME_LEN of
+            true -> binary:part(Name, 0, ?MAX_USERNAME_LEN);
+            false -> Name
+        end,
+    <<<<C>> || <<C>> <= Capped, C >= 32, C =/= 127>>.
```
lukebakken: **Uncaught crash: this can emit invalid UTF-8, which makes the JSON encoder raise and converts a 422 into HTTP 500.**

```erlang
<<<<C>> || <<C>> <= Capped, C >= 32, C =/= 127>>.
```

Both the cap and the filter are byte-oriented, so neither preserves UTF-8 validity. Two reachable inputs, both of which I reproduced against the resolved deps:

`san_type = ip`. `rabbit_cert_info` returns an `iPAddress` SAN as a raw 4-byte binary, and `extract_username/2` passes it through unchanged, so `Username = <<192,168,1,10>>`. The filter drops the `1` and `10` as control bytes but keeps 192 and 168, which are not valid UTF-8 on their own:

```erlang
> sanitize_username(<<192,168,1,10>>).
<<192,168>>
> rabbit_json:encode(#{<<"reason">> => <<"no internal user named ", 192,168, " exists">>}).
** exception error: {invalid_byte,<<"0xC0">>, ...}
```

A CN or DN longer than 256 bytes whose 256th byte splits a multibyte character. `binary:part(Name, 0, 256)` leaves a dangling lead byte:

```erlang
> rabbit_json:encode(#{<<"reason">> => sanitize_username(<<255 bytes, 16#C3, 16#A9>>)}).
** exception error: {invalid_byte,<<"0xC3">>, ...}
```

CN values are not hex-escaped by `rabbit_cert_info`, unlike DN values, so the second case is reachable in `common_name` mode with an ordinary long non-ASCII CN.

`reply_error` runs inside `with_semaphore/7`'s catch-all, so in both cases the raise is swallowed and the operator gets HTTP 500 `internal_error` instead of the 422 `auth_failed` the code intends, losing the diagnostic entirely. Truncating on a character boundary and rejecting or escaping non-UTF-8 bytes would fix both. The `ip` case also raises a separate question: is a raw address binary a sensible username to look up, or should `san_type = ip` format it first?

lukebakken (changes requested):
Normal-effort review of 9970e91. Eight findings, all verified empirically rather than by reading alone: real openssl certificate fixtures, live `ssl:listen` and `ssl:connect` handshakes for every broker-divergence claim, and direct calls into the deps this checkout resolves (OTP 27.3.4.14, public_key 1.17.1.4, ssl 11.2.12.10). Inline comments carry the reproductions.

First, credit where it is due: 9970e91 is a real bug fix. The leaf-first chain was being handed to `pkix_path_validation/3`, which wants anchor-closest-first, and every pre-existing test used a single-cert direct CA-to-leaf chain, which is exactly the shape that masks it. Layer 2's deliberate strictness over upstream (treating an empty DN as extraction failure) is well argued in the comment, and I agree with it for a validation endpoint.

The theme of the findings is one thing, not eight. This endpoint predicts whether TLS material will work on a live listener, and seven of the eight are places where its verdict diverges from what the broker actually does. Ranked by severity:

1. Intermediate-only CA bundle returns 204 "usable" while a real `verify_peer` listener rejects every client with `unknown_ca` (line 619). This is the only **false PASS**, which makes it the worst of the set: an operator who omits the root is told their material is fine and then finds nothing can connect.
2. CA key rollover bundle flips verdict on PEM order (line 626).
3. `otherName` SAN raises `badarg`, giving HTTP 500 with no reason (line 683).
4. `sanitize_username/1` can emit invalid UTF-8, so the JSON encoder raises and a 422 becomes HTTP 500 (line 747).
5. `resolve_user/1` consults only the internal backend, so LDAP-backed or OAuth2-backed brokers get a false "no internal user named X exists" (line 721).
6. Depth violations are reported as "does not chain", which is actively false (line 608).
7. Leaf-first-but-unordered chains are rejected where the broker rebuilds and accepts (line 597).
8. Dead `{bad_cert, {cert_expired, _}}` clause, provable by dialyzer (line 606). Cleanup only.

Findings 3 and 4 are the two that produce HTTP 500 with no reason at all, both from caller-supplied certificate content, so they break the backend contract that every outcome carries a category. They are also the cheapest to fix.

Three things I want to explicitly clear, because they came up during review and I do not want them re-litigated:

- `check_verify_for_cert_login/1`'s `#{ssl_options := SslMap}` head is safe. `parse_ssl_options/2` runs earlier in the pipeline and always inserts the key, so no `function_clause` is reachable.
- `{max_path_length, N}` is exactly what `ssl` derives from `depth` (`ssl_handshake.erl:2138`), with no off-by-one. I verified depth 0 rejects leaf+intermediate and accepts a directly-signed leaf.
- `'EncryptedPrivateKeyInfo'` in `?PRIVATE_KEY_TYPES` is dead rather than divergent, so the R6 private-key rejection is intact. OTP rewrites that tag to `'PrivateKeyInfo'` inside `pubkey_pem:decode_encrypted_private_keyinfo/1` before a caller ever sees it, which I confirmed with both PKCS#8 and traditional AES-encrypted PEMs.

Two smaller items that fell below the cut but are cheap while you are in here:

- OpenSSH-format private keys decode as `{no_asn1, new_openssh}`, which none of the private-key type lists covers, so such a key concatenated with a certificate passes the R6 rejection in `classify_pem_entries/2` and the request is accepted.
- `parse_cert_login_from/2` accepts `subject_alt_name` as an alias (with a dedicated test, so clearly intentional), but `?REASON_BAD_CERT_LOGIN_FROM` names only three values, so the endpoint's own error message contradicts what it accepts. Note the alias is not legacy here: both spellings landed together upstream in 13b2b0f522 with `subject_alternative_name` as the documented one, and this HTTP field is new, so there is no released client to keep compatible. Either mention it in the reason string or drop the clause.
- The CA PEM is decoded twice per `client_cert` request (line 303 inside a `try`, line 566 outside one), and `find_by_subject/2` re-decodes every bundle CA that `check_cert_validity/1` already decoded. Threading the decoded DERs through would remove both and make the `skip` clause on line 581 provably dead, so it and the `| skip` half of the spec could go. Efficiency and clarity only; I confirmed there is no input that survives the first decode and then throws at the second.

On coverage: CI is green on all five checks, but none of the seven divergences has a test, and the assertion that would most likely have caught one (the depth case) is the single wildcard reason in the PR. The gap is structural rather than accidental: the suite tests this module against `public_key` while the thing being predicted is `ssl`'s path building, and those disagree in all the ways above. A few cases that assert the endpoint's verdict matches a real handshake with the same material would be worth more than more `public_key`-only cases, and would have caught findings 1, 2 and 7.

Requesting changes on the strength of finding 1: a false PASS on a bundle that cannot serve a single connection is the failure mode this endpoint exists to prevent, and shipping it would leave operators worse off than with no validation at all. Findings 3 and 4 also matter before merge, since both let caller-supplied certificate content produce an HTTP 500 with no diagnostic, breaking the contract that every outcome carries a category. They are also the cheapest of the set to fix.

The rest are not merge blockers from my side. Findings 2, 5, 6 and 7 could reasonably ship as documented known limitations if you would rather land the layers now and close the broker-parity gaps in a follow-up; if you go that route, please say so explicitly in the comments so the divergences are recorded rather than latent. Finding 8 and the three smaller items are cleanup at your discretion.

src/aws_auth_validate_tls.erl
sdewhitt: Fixed in `eaacfc4`. `find_all_self_signed_by_subject/2` now requires `public_key:pkix_is_self_signed/1`, so an intermediate-only bundle finds no anchor and fails. Parity tests in `7c76440` pin bundle=[Int] rejected for both `[leaf]` and `[leaf, int]` against a real handshake.

Your fourth control case (`bundle=[Root, Int]`, client `[leaf]` -> handshake ok) caught a regression this introduced: requiring a self-signed anchor also stopped bundle intermediates being used to *extend* the client chain, turning that into a false FAIL. `d64a662` adds `extend_with_bundle_intermediates/2`, mirroring `ssl_certificate:certificate_chain/3`, and the control is now a pinned parity test.

src/aws_auth_validate_tls.erl
sdewhitt: Fixed in `eaacfc4`. `find_trust_anchors/2` collects every self-signed subject-DN match and `try_anchors/3` accepts if any validates, so the verdict no longer depends on concatenation order. Both bundle orders are pinned as parity tests against a real handshake in `7c76440`.

src/aws_auth_validate_tls.erl
sdewhitt: Fixed in `ee3bbfc`. The `{error, _}` return from `rabbit_cert_info:sanitize_other_name/1` is now matched explicitly and returns `auth_failed` / `?REASON_EXTRACT_FAILED`. The clause structure also routes any non-binary/non-iolist return to a categorized failure rather than `iolist_to_binary/1`, so no shape reaches the handler catch-all. Test mocks the error tuple and asserts no raise.

src/aws_auth_validate_tls.erl
sdewhitt: Fixed in `ee3bbfc`, then bounded in `b1e7ca0`. Truncation is on a character boundary and invalid bytes are hex-escaped, so the output is always valid UTF-8.

Review of that fix caught a second problem: escaping ran after truncation, so an all-invalid input expanded ~6x past the cap (300 bytes of `0xC0` returned 1530). `b1e7ca0` re-truncates post-escape. Tests assert a real `rabbit_json:encode/1` round-trip rather than just inspecting bytes, since encodability was the actual failure mode.

src/aws_auth_validate_tls.erl
sdewhitt: Fixed in `ee3bbfc`. The inaccurate non-divergence comment is gone. When any non-internal authN backend is configured, a username absent from the internal table now returns `config_conflict` with an honest inconclusive reason instead of a confident `auth_failed`; when internal is the only configured backend the verdict is unchanged.

Deliberately did not call `check_user_login/2` or `exists/1` on external backends: both can perform network I/O (LDAP bind, HTTP request), which would violate R3 and add an unbounded outbound call to a currently-pure path. `only_internal_authn_configured/0` reads the `auth_backends` env only, handling both bare-module and `{AuthN, AuthZ}` entries.

Flagging the contract change: this is a new `config_conflict` case on a path that previously always returned `auth_failed`.

src/aws_auth_validate_tls.erl
sdewhitt: Fixed in `eaacfc4`. `max_path_length_reached` now maps to a dedicated `?REASON_CHAIN_DEPTH_EXCEEDED`, and `invalid_signature`, `invalid_issuer`, `invalid_key_usage` and `missing_basic_constraint` each get their own accurate reason via `pick_best_failure/1`.

Went with fixed per-atom macros rather than interpolating the atom into the string, to keep the response surface fixed-category per R4. `?REASON_CHAIN_FAILED` remains the fallback for unknown reasons.

Also tightened `tls_validate_chain_intermediate_depth_zero_rejected_test` to pin the depth reason instead of wildcarding it, as you noted.

src/aws_auth_validate_tls.erl
sdewhitt: Fixed in `eaacfc4`. `order_chain/1` rebuilds the tail by issuer/subject linkage with the leaf pinned at element 0, so `[leaf, root, int]` now validates.

Your narrowing is preserved: a fully root-first chain is still rejected, and `tls_validate_chain_root_first_rejected_test` plus a parity test pin that both sides reject it, so the reorder cannot be loosened later. Also updated `?REASON_BAD_CLIENT_CERT` to state the leaf-first requirement, since it was previously only in implementation comments.

src/aws_auth_validate_tls.erl
sdewhitt: Removed in `eaacfc4`. The bare-atom clause and the catch-all cover everything.

Related, from S1: `gmake dialyze` also flagged the `{no_asn1, _}` clause in `is_known_key_type/1` as unreachable, but that one is a false positive -- `public_key`s exported `pem_entry()` enumerates only ASN.1 record tags while `pubkey_pem` really does emit `{no_asn1, new_openssh}` (verified at runtime). `63eccad` scopes a `nowarn_function` to that function with the evidence in a comment, so dialyzer stays quiet over the module without deleting a live clause.

src/aws_auth_validate_tls.erl
lukebakken: Verified fixed at ced9d0f. `order_chain/1` plus `order_chain_loop/3` rebuild the tail by issuer/subject linkage with the leaf pinned at element 0.

Confirmed independently rather than by reading: a differential harness of my own comparing the endpoint verdict against a real `ssl:listen`/`ssl:connect` handshake now agrees on `[leaf, root, int]` (both accept), and the narrowing holds, root-first paste is still rejected on both sides.

I also mutation-tested the coverage, since a test that cannot fail is not coverage. Replacing `order_chain(ClientDers)` with `ClientDers` fails `unordered_tail_leaf_root_int_both_accept` with the exact divergence I originally reported: `ssl=accept, endpoint=reject`. So the fix is genuinely pinned.

The leaf-first requirement is now in `?REASON_CLIENT_CERT_INVALID` ("leaf certificate first"), which was the documentation half of this comment.

src/aws_auth_validate_tls.erl
lukebakken: Verified fixed at ced9d0f. The tuple clause is gone and `gmake dialyze` over this component is now clean for `aws_auth_validate_tls`: zero warnings attributed to the module. The remaining warnings in the run are all in files this PR does not touch (`aws_auth_validate_ldap`, `aws_auth_validate_mgmt`, and dependency modules).

Agreed on the `{no_asn1, new_openssh}` call, and I checked it rather than taking it on trust: `pubkey_pem` does emit that tag at runtime while the exported `pem_entry()` type does not admit it, so this is a gap in the published spec and not dead code. Scoping `nowarn_function` to the one function with the evidence in a comment is the right resolution; deleting the clause would have reopened the OpenSSH bypass.

src/aws_auth_validate_tls.erl
lukebakken: Verified fixed at ced9d0f. `pick_best_failure/1` maps `max_path_length_reached` to a dedicated reason and gives `invalid_signature`, `invalid_issuer`, `invalid_key_usage` and `missing_basic_constraint` their own strings, with `?REASON_CHAIN_FAILED` as the fallback.

Confirmed the diagnosis is now correct on the case from my comment: depth 0 with a leaf+intermediate chain returns "the client certificate chain exceeds the configured path length (depth)", and a wrong-CA bundle returns the distinct invalid-signature reason rather than the same string.

No argument on fixed per-atom macros over interpolating the atom. A closed set of reasons is the better contract for a diagnostic API, and it keeps the atom names from becoming a de facto public interface.

The tightened assertion does its job: dropping the `max_path_length_reached` line from `pick_best_failure/1` so it falls through to the generic reason fails `tls_validate_chain_intermediate_depth_zero_rejected_test` on the pinned string. Under the original wildcard that mutation would have passed silently, which was the point of the note.

src/aws_auth_validate_tls.erl
lukebakken: Verified fixed at ced9d0f. This was the finding I requested changes on, and it is properly closed.

My own differential harness confirms the false PASS is gone: `bundle=[Int]` now fails for both `[leaf]` and `[leaf, int]`, matching the `unknown_ca` a real listener returns, and all four of my original cases plus six more agree with a live handshake.

Good catch on the regression, and it is the part of this fix I most wanted to check. `extend_with_bundle_intermediates/2` correctly mirrors cacerts being chain-building material and not only anchors, and the recursion is safely bounded since each appended CA is removed from the candidate pool, so a cross-signed pair cannot loop.

Worth recording that the two legs are independently pinned. Disabling the bundle-intermediate extension fails `root_int_bundle_leaf_alone_both_accept`; the self-signed anchor requirement is what carries the int-only cases. One note for the record: removing `pkix_is_self_signed` from `find_all_self_signed_by_subject/2` alone does not break any test, but that is because it is unreachable rather than untested. `extend_with_bundle_intermediates/2` runs first and consumes every non-self-signed bundle CA matching the top cert's issuer, so no non-self-signed candidate can survive to the anchor search. The guard is correct to keep as defence in depth.

src/aws_auth_validate_tls.erl
lukebakken: Verified fixed at ced9d0f. `find_trust_anchors/2` collects every self-signed subject-DN match and `try_anchors/3` accepts if any validates.

Confirmed order independence against a live handshake with the rollover twins from my comment (both `CN=TestRoot`, different keys, leaf chaining to the older): both `[Old, New]` and `[New, Old]` now pass on both sides, where previously only one order did.

Mutating `try_anchors/3` to try only `hd(Anchors)` fails `key_rollover_old_new_order_both_accept` with exactly the original symptom, `invalid signature` against the wrong twin, so the retry cannot be silently lost later.

src/aws_auth_validate_tls.erl
lukebakken: Verified fixed at ced9d0f. The `{error, _}` return is matched explicitly and yields `auth_failed` / `?REASON_EXTRACT_FAILED`, and the surrounding clauses route any non-binary/non-iolist shape to a categorized failure instead of `iolist_to_binary/1`.

Checked the coverage by reverting the clause block to the original `{ok, iolist_to_binary([Value])}`: `tls_extract_other_name_error_tuple_test` then fails with `**error:badarg`, the precise crash from my comment. So the test genuinely reproduces the HTTP 500 rather than just asserting the new return shape.

One observation, not a request: `{error, _}` is a tuple, so the `_Other` clause below would catch it and return the same categorized reason even without the explicit clause. Keeping both is fine, the named clause carries the reasoning for the next reader.

src/aws_auth_validate_tls.erl
lukebakken: Verified fixed at ced9d0f. The inaccurate non-divergence comment is gone and `resolve_user/1` is now three-way: internal-only configuration keeps the confident `auth_failed`, and a username absent from the internal table with any non-internal authN backend configured returns `config_conflict` with an inconclusive reason. `only_internal_authn_configured/0` handles both bare-module and `{AuthN, AuthZ}` entries, which is the shape `rabbit_access_control` iterates.

I agree with not calling `check_user_login/2` or `exists/1` on external backends, and I would have pushed back if you had. An LDAP bind or outbound HTTP request from a validation endpoint would trade a false FAIL for an unbounded blocking call, which is a worse defect than the one being fixed. Reading the env and reporting honestly that the answer is inconclusive is the right resolution: it tells the operator what was actually checked, which was the real gap.

Thanks for flagging the contract change explicitly. Agreed that is the notable consequence here: a path that previously only ever returned `auth_failed` can now return `config_conflict`, so anything treating a non-`auth_failed` category as unexpected will see a new case.

src/aws_auth_validate_tls.erl
lukebakken: Verified fixed at ced9d0f. Truncation is on a character boundary, invalid bytes are hex-escaped as `<0xHH>`, and the output is re-truncated after escaping.

Confirmed against the failing inputs from my comment: `<<192,168,1,10>>` and a 255-byte CN split mid-codepoint both now `rabbit_json:encode/1` cleanly, along with a dozen other edge cases I threw at it, and a 1000-byte all-invalid input comes back within the 256-byte cap.

Good catch on the ~6x expansion in review of your own fix; that was a real second bug and the ordering is now right. Reverting the sanitizer to the byte-oriented cap without escaping fails three tests, including the raw-IP and split-multibyte cases, so both the validity and the bound are pinned.

Asserting a real `rabbit_json:encode/1` round-trip rather than inspecting bytes is the right call: encodability was the actual failure mode, and a byte-level assertion would have passed on some inputs that still raise.

lukebakken (approved):
Approving at ced9d0f. All eight round-1 threads are resolved, CI is green across all four broker series plus format-check, and `gmake dialyze` is clean for `aws_auth_validate_tls` (the remaining warnings in the run are all in files this PR does not touch).

The finding I requested changes on is properly closed. My own differential harness, comparing the endpoint verdict against a real `ssl:listen`/`ssl:connect` handshake, now agrees on all ten cases I ran: the intermediate-only bundle fails on both sides for `[leaf]` and `[leaf, int]`, both rollover-twin bundle orders pass on both sides, the unordered tail passes on both, and root-first paste is still rejected on both.

I mutation-tested the coverage rather than treating green as proof, since a test that cannot fail is not coverage. Reverting each fix in turn reproduces the exact symptom originally reported:

```
order_chain disabled                -> unordered_tail_leaf_root_int_both_accept fails (ssl=accept, endpoint=reject)
bundle-intermediate ext disabled    -> root_int_bundle_leaf_alone_both_accept fails
only hd(Anchors) tried              -> key_rollover_old_new_order_both_accept fails (invalid signature)
iolist_to_binary([Value]) restored  -> tls_extract_other_name_error_tuple_test fails (**error:badarg)
byte-cap, no UTF-8 escaping         -> 3 tests fail, incl. raw-IP and split-multibyte
depth reason mapping removed        -> depth_zero_rejected fails on the pinned reason
```

The parity suite is the most valuable thing in this PR. It closes the structural gap I raised, testing against `ssl`'s path building rather than `public_key`'s, and `write_chain_file/1` using a unique filename per case avoids the PEM-path caching trap that silently invalidates naive versions of this test (it produced false results in my own first harness before I caught it).

Two items recorded on the threads rather than left latent:

- Removing `pkix_is_self_signed/1` from `find_all_self_signed_by_subject/2` breaks no test, but because it is unreachable rather than untested: `extend_with_bundle_intermediates/2` runs first and consumes every non-self-signed bundle CA matching the top cert's issuer. Correct to keep as defence in depth.
- `resolve_user/1` can now return `config_conflict` on a path that previously only ever returned `auth_failed`. The author flagged this; anything treating a non-`auth_failed` category as unexpected will see a new case. Not calling `check_user_login/2` or `exists/1` on external backends is the right call, an LDAP bind or outbound HTTP request from a validation endpoint would trade a false FAIL for an unbounded blocking call.

Nice work on the two regressions caught while fixing: the bundle-intermediate false FAIL introduced by the anchor requirement, and the ~6x expansion from escaping after truncation. Both were found by reviewing your own fixes, which is where they are cheapest to find.

PR #167, opened 2026-07-30
Auth Validation: Name the failing fields in ARN resolution errors

sdewhitt:
_Enhancement to #43_

The generic "failed to resolve ARN" gave an operator with several ARNs in one request no way to tell which one was broken, so they had to bisect by editing ARNs and retrying -- the guess-and-retry loop this endpoint exists to remove. 

Responses now name the fields like:
"ARN resolution failed for: password_arn, ssl_options.cacertfile_arn"

Changes:
  - add aws_auth_validate_ssl:arn_resolve_reason/1 to render the list, falling back to the old wording when no attribution is available
  - resolve ALL referenced ARNs instead of short-circuiting on the first failure, so one response reports every broken field; bounded at three ARNs per request and already gated by the assume_role guardrail and the concurrency semaphore
  - attribute failures structurally ({arn_failed, Fields}) rather than by re-parsing reason strings; add resolve_ssl_material/2 for the shared http/oauth path and aggregate_arn_results/2 for ldap
  - resolve the mTLS cert and key ARNs before decoding either, so a request with both broken names both fields
  - distinguish CONTENT failures from RESOLUTION failures: a certfile or keyfile ARN that resolves but holds no usable material now says so instead of claiming resolution failed
  - qualify nested keys (ssl_options.*) so the operator knows where in the request to look
  - drop the now-dead per-field LDAP resolve macros

Field names come from the caller's own request, so they disclose nothing new. The resolved content and the underlying AWS error are still never included, keeping "does not exist" and "access denied" indistinguishable. A password_arn shape error still short circuits, preserving ARN-first ordering: a malformed request triggers no secret fetch.

### DIFF
--- a/src/aws_auth_validate_http.erl
+++ b/src/aws_auth_validate_http.erl
@@ -728,29 +728,22 @@ build_client_ssl_opts(#{ssl_options := Map} = Params) ->
     %% instance role. Default to `none' (never a usable state) if the key is
     %% somehow absent, preserving the fail-closed contract.
     State = maps:get(aws_state, Params, none),
-    case
-        aws_auth_validate_ssl:resolve_cacerts(maps:get(<<"cacertfile_arn">>, Map, undefined), State)
-    of
+    %% Resolves the CA bundle and the mTLS pair together so a response can name
+    %% every ARN field that failed, not just the first.
+    case aws_auth_validate_ssl:resolve_ssl_material(Map, State) of
         {error, _, _} = Err ->
             Err;
-        {ok, CacertOpts} ->
-            case aws_auth_validate_ssl:resolve_client_cert(Map, State) of
-                {error, _, _} = Err ->
-                    Err;
-                {ok, ClientOpts} ->
-                    Opts =
-                        CacertOpts ++ ClientOpts ++
-                            aws_auth_validate_ssl:translate_ssl_opts(Map, <<"sni">>),
-                    %% Whether the caller set verify explicitly governs the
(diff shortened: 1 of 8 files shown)


lukebakken (approved):
> [!NOTE]
> This review was written by Claude (Anthropic's Claude Code) under the direction of @lukebakken, who reviewed it before posting. The findings and verification results were produced by Claude running the tests, dialyzer, and erlfmt described below.

Reviewed the ARN field-attribution work and the follow-up cleanup in 288f8c8.

The aggregation design is sound: resolution failures collect across every ARN in a request and name each failing field in one response, while a content failure (ARN resolved but the payload is unusable) is reported as-is. The ARN-first ordering invariant holds, so a malformed request triggers no secret fetch. Field names echo the caller's own request keys and never the resolved bytes, and the underlying AWS error is dropped, keeping the response categories fixed.

The follow-up commit removes a bare atom that had leaked into the reason position from `resolve_cacerts/2`, drops a dialyzer-flagged unmatchable clause in the LDAP aggregator, and refreshes two stale comments.

Verification:

- eunit: aws_auth_validate_tests 182, aws_auth_validate_http_tests 94, aws_auth_validate_tls_tests 95, all passing
- Common Test: ldap 18, tls 11, http 14, oauth 12, mgmt 20, zero failures (3 mgmt cases skip on the OAuth-authz availability gate, unrelated to this change)
- dialyzer: no warnings for the changed modules; the remaining warnings are in code byte-identical to main and pre-date this PR
- erlfmt clean

LGTM.

PR #168, opened 2026-07-30
Comment cleanup: inline the design-doc requirement IDs

sdewhitt:
> [!NOTE]
> This description was drafted by Claude (Anthropic's Claude Code) under the direction of @lukebakken, who reviewed it before posting.

Lands after #167. Cleans up comments across the auth validation endpoint code, replacing internal design-doc requirement IDs (R3/R4/R6, etc.) with inline rationale so the reasoning is legible to any contributor, and restoring two same-repo GitHub references (#116, #153) that resolve for any reader.

Also updates an aws_lib_config test fixture to use AWS documentation example account IDs. Comments only, apart from that fixture change.

### DIFF
--- a/src/aws_auth_validate_backend.erl
+++ b/src/aws_auth_validate_backend.erl
@@ -23,7 +23,9 @@
     %% config mismatch (token_invalid -- the JWKS the broker fetches will reject
     %% live tokens). Safe to be granular here: unlike the reachability
     %% categories, these describe the caller's own token, not the broker's infra
-    %% or an SSRF target, so they leak nothing R4 is guarding.
+    %% or an SSRF target, so they disclose nothing the fixed-category rule guards
+    %% (that rule prevents echoing raw errors, hostnames, or network details that
+    %% would enable SSRF reconnaissance).
     | token_expired
     | token_invalid.
 
--- a/src/aws_auth_validate_http.erl
+++ b/src/aws_auth_validate_http.erl
@@ -728,29 +728,22 @@ build_client_ssl_opts(#{ssl_options := Map} = Params) ->
     %% instance role. Default to `none' (never a usable state) if the key is
     %% somehow absent, preserving the fail-closed contract.
     State = maps:get(aws_state, Params, none),
-    case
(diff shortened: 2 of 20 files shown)


lukebakken (approved):
> [!NOTE]
> This review was written by Claude (Anthropic's Claude Code) under the direction of @lukebakken, who reviewed it before posting. The verification results below were produced by Claude running the tests and erlfmt described.

Reviewed the comment cleanup at 90cabab.

Replacing the internal requirement IDs with inline rationale is a clear improvement: an unlookupable label like `R4` pointed at reasoning no external contributor could read, and the rewrites state the actual rule instead. I checked every rewritten comment against the code it documents and found no case where the new wording misdescribes behavior. The security-relevant claims (ARN-first ordering, no credential in a crash report, verify_peer never silently downgraded, decision parity with the upstream broker backend) are preserved intact, including the LOAD-BEARING markers.

Three follow-ups were addressed in 90cabab: the sweep had left five requirement-ID references in `aws_auth_validate_tls.erl` and its test, and had dropped the `#116` and `#153` references, which resolve for any reader of this repo and match the convention 2942798's own message says it preserves. Those are restored.

The `aws_lib_config` fixture now uses documentation example values. Worth noting for the record: the previous value remains in this repo's history, so the fixture change sanitizes the tip only. An AWS account ID is not a credential and is not treated as a secret, so this is hygiene rather than a vulnerability, and it does not warrant rewriting public history.

Verification at 90cabab:

- the diff is comment-only apart from the fixture strings, confirmed mechanically (no changed line outside a `%%` comment)
- eunit: aws_auth_validate_tests 182, aws_auth_validate_http_tests 94, aws_auth_validate_tls_tests 95, aws_lib_config 70, all passing
- Common Test: tls 11, ldap 18, mgmt 20, zero failures (3 mgmt cases skip on the OAuth-authz availability gate, unrelated to this change)
- erlfmt clean
- no remaining requirement-ID references in `src/` or `test/`

CI was still pending when this was written; the local erlfmt run covers what `format-check` gates.

LGTM.

PR #170, opened 2026-08-03
`aws_lib`: Remove the unused two-step connection API

sdewhitt:
Fixes #108.

## Change

Removes the two-step connection API from `aws_lib`, which had no callers anywhere:

- `open_connection/2`, `open_connection/3`
- `direct_request/7`
- `close_connection/1`
- the `connection_handle()` type, now unused
- `create_uri/2` and `direct_gun_request/6`, reachable only from the above

147 lines from `src/aws_lib.erl`, plus the doc references in `AGENTS-aws_lib.md`
and `README-aws_lib.md`.

Per the issue's sequencing note and @lukebakken's comment: this is the speculative,
untested, retry-bypassing version. If the reuse work later wants a
hold-a-connection-across-requests primitive, it should be designed and tested
deliberately there rather than adopted from here.

## Evidence the removal is safe

- No callers in `src/` or `test/`, and none in the sibling worktrees.
- No dynamic references (`apply/3`, atom-form calls, meck expectations,
  `-ifdef(TEST)` exports).
- `aws_lib_httpc:open/3` and `close/1` retain other callers, so they are not
  orphaned by this.
- The surviving reuse machinery is untouched and confirmed intact:
  `enable_connection_reuse/1`, `close_reuse_connection/1`, `seed_conn_from_state/1`,
  `finish_conn/2`, `perform_request_reuse/8`.

## Verification

- eunit: **814 passed, 0 failed** (matches the `main` baseline)
- `erlang:function_exported/3` returns `false` for the removed functions;
  `aws_lib:module_info(exports)` no longer lists them; calling one raises `undef`
- the surviving request path still works end to end against a local HTTP listener,
  including the `enable_connection_reuse/1` path
- no new dialyzer warnings (the ~30-warning backlog in #8 is pre-existing and
  untouched here)

### DIFF
--- a/AGENTS-aws_lib.md
+++ b/AGENTS-aws_lib.md
@@ -40,9 +40,8 @@ Checked in order by `aws_lib_config:credentials/1`:
 
 ### HTTP Client
 
-Uses Gun (HTTP/1.1 and HTTP/2). Two usage patterns:
-- **One-shot:** `request/6-8` opens connection, makes request, closes connection
-- **Pooled:** `open_connection/2` + `direct_request/7` + `close_connection/1`
+Uses Gun (HTTP/1.1 and HTTP/2). One-shot request lifecycle:
+- `request/6-8` opens connection, makes request, closes connection
 
 ### Retry Logic
 
@@ -52,7 +51,7 @@ Uses Gun (HTTP/1.1 and HTTP/2). Two usage patterns:
 
 | Module | Purpose |
 |--------|---------|
-| `aws_lib` | Main API: state management, request lifecycle, retries, connection pooling, EBS volume discovery |
+| `aws_lib` | Main API: state management, request lifecycle, retries, EBS volume discovery |
 | `aws_lib_config` | Credential/region loading from env, files, and IMDS; INI parsing; IMDSv2 token management |
 | `aws_lib_sign` | AWS Signature Version 4: canonical request, signing key derivation, authorization header |
 | `aws_lib_ops` | Higher-level EC2 operations: create/delete EBS snapshots |
--- a/README-aws_lib.md
+++ b/README-aws_lib.md
@@ -69,14 +69,6 @@ RequestHeaders = [{"Content-Type", "application/x-amz-json-1.0"},
 | `aws_lib:api_get_request/3` | GET with automatic retries |
 | `aws_lib:api_post_request/5` | POST with automatic retries |
 
-### Connection Pooling
-
-| Function | Description |
-|----------|-------------|
-| `aws_lib:open_connection/2-3` | Open a reusable connection to a service |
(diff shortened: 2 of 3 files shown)


lukebakken (approved):
> [!NOTE]

Reviewed the dead-code removal at b06453d.

For a removal, the question is whether the code was truly unreachable and whether anything still needed went with it. Both check out:

- **Genuinely dead before removal.** At the merge base, the only references to `open_connection/2,3`, `close_connection/1`, `direct_request/7`, `create_uri/2`, `direct_gun_request/6`, and the `connection_handle()` type were internal to that cluster: `open_connection/2` called `/3`, and `direct_request/7` called `create_uri/2` and `direct_gun_request/6`. Nothing outside the cluster reached into it.
- **No callers anywhere.** A sweep across the umbrella found only unrelated same-named functions in other modules (`amqp10_client:open_connection`, `rabbit_mgmt_util:direct_request`), with zero `aws_lib:` call sites. `direct_gun_request` and `create_uri` have no matches at all outside `aws_lib`. No dynamic references either (no `apply(aws_lib, ...)`, no `{aws_lib, ...}` tuple form).
- **The reuse path is intact.** `enable_connection_reuse/1`, `close_reuse_connection/1`, `seed_conn_from_state/1`, `finish_conn/2`, and `perform_request_reuse/8` are all still present and still referenced, so the supported connection-reuse pattern is unaffected. The shared helpers the removed code happened to use (`aws_lib_httpc:open/close`, `endpoint_host/1`, `gun_open_opts/1`) retain other live callers, so nothing is left orphaned.
- **Docs match the code.** `AGENTS-aws_lib.md` drops the "Two usage patterns" and pooling references and `README-aws_lib.md` drops the "Connection Pooling" table. No stale mention of the removed API survives in `src/` or the markdown files.

Verification: the plugin compiles clean, and the compiled beam confirms none of the six removed functions are still exported while the surviving API (`enable_connection_reuse`, `close_reuse_connection`, `api_get_request`, `api_post_request`) still is. `aws_lib_tests` passes 50/0 locally and the full eunit run passes with zero failures. All four `build-test` jobs and `format-check` are green.

Two notes for the record rather than objections:

- My local full eunit run reported 805 passing where the description cites 814. Zero failures either way, and CI is green on all four server versions, so this looks like environment (no live AWS metadata endpoint here) rather than a regression. I did not reproduce 814 exactly.
- Dialyzer reports 12 warnings, all in modules this PR does not touch (`aws_auth_validate_*`, `aws_lib_config.erl`, `aws_sup.erl`). None reference `aws_lib.erl`, the only changed `.erl` file, so the "no new warnings" claim holds by attribution. I confirmed it that way rather than by diffing against a fresh `main` PLT run.

LGTM.

PR #174, opened 2026-08-03
`aws_lib`: Replace the fixed retry delay with exponential backoff and jitter

sdewhitt:
Fixes #81.

## Change

Replaces the fixed 500ms `?LINEAR_BACK_OFF_MILLIS` delay with **equal-jitter**
exponential backoff:

```erlang
backoff_delay(Attempt, Base, Cap) ->
    Temp = min(Cap, Base * (1 bsl Attempt)),
    HalfTemp = max(1, Temp div 2),
    HalfTemp + rand:uniform(HalfTemp).
```

`?LINEAR_BACK_OFF_MILLIS` becomes `?BACKOFF_BASE_MILLIS` (500) and
`?BACKOFF_CAP_MILLIS` (10000). An explicit `Attempt` counter is threaded through
the retry loop, since the loop counts retries *down* and the exponent needs the
ordinal.

**Equal-jitter over erlcloud's full-jitter**: full jitter can produce a ~1ms sleep,
which defeats the purpose of backing off. Equal jitter guarantees a `Temp/2` floor
while still spreading concurrent retriers. `1 bsl Attempt` rather than
`math:pow/2` avoids float conversion and a `badarith` at large exponents (raised
in review). `max(1, ...)` keeps `rand:uniform/1` from ever seeing 0.

Issue #80's non-retriable short-circuit is untouched -- only `retriable` verdicts
reach the delay.

## Verification

A second commit stubs `timer:sleep` in the two retry-behaviour eunit
generators (the passthrough+unstick meck pattern the suite already uses for
`calendar`). Exponential backoff makes a full 5-retry exhaustion sleep for
several seconds, which tripped eunit's 5s per-test default under `make tests`
and surfaced as *timed out*. Those cases assert retry behaviour, not delay
timing, so the no-op sleep changes nothing they check; the delay arithmetic is
covered by `backoff_delay/3`'s own tests.

eunit **821 passed, 0 failed** (baseline `main` is 805). 16 new tests: 12 for
`backoff_delay/3` covering ranges, boundaries, jitter, monotonicity, and
`Cap < Base`, plus 4 for `maybe_backoff/3` and the sleep-count invariant.

Observed growth, then jitter across repeated calls with identical arguments:

```
Attempt 0: 424ms   Attempt 3: 3654ms
Attempt 1: 765ms   Attempt 4: 4667ms
Attempt 2: 1937ms  Attempt 6: 8973ms   (capped)

same args, 5 calls: 1905, 1531, 1427, 1918, 1183 ms
```

Driven through the **real retry path** against a local endpoint returning 500:

```
5 attempts, 4 sleeps: 257ms -> 767ms -> 1722ms -> 3773ms   (6519ms total)
```

Three full runs differ (6519ms / 6456ms / 5694ms total), confirming the jitter
is live rather than a fixed schedule. Every run performs exactly one fewer sleep
than attempt: the delay after the final attempt is skipped, since it would push
back the returned error without buying another attempt. Boundary cases
`backoff_delay(0,0,0)`, `(0,1,1)` and `Cap < Base` all return sane positive
values without crashing.

## Latency tradeoff for reviewers

Worst case cumulative delay rises from **2.5s** (5 x 500ms on `main`) to
**7.5s** (500+1000+2000+4000 max, expected ~5.6s). Only the gaps *between*
attempts sleep now, so 5 attempts incur 4 delays rather than 5; keeping the old
fixed delay under that rule would have cost 2.0s. This sits on the broker-boot
ARN resolution path, so it is a real change in worst-case boot time and deserves
an explicit decision. The cap bounds it; lower `?BACKOFF_CAP_MILLIS` if 7.5s is
too long for boot.

## Merge order

Please merge this **before** #85 (retry decoupling). Both touch
`api_request_with_retries`, and #85 adds a `wait_time_ms` seam that this policy
plugs into -- landing the policy first and then extracting around it is the
cleaner direction. `?BACKOFF_CAP_MILLIS` is intentionally applied at the call site
here rather than threaded as a parameter, since making retry options
caller-configurable is #85's scope.

### DIFF
--- a/AGENTS-aws_lib.md
+++ b/AGENTS-aws_lib.md
@@ -45,7 +45,7 @@ Uses Gun (HTTP/1.1 and HTTP/2). One-shot request lifecycle:
 
 ### Retry Logic
 
-`api_get_request/3` and `api_post_request/5` retry up to 5 times with a fixed 500ms delay between attempts. Credentials are validated before the first attempt and re-validated before each retry. If credentials cannot be loaded, returns `{error, {credentials, Reason}}` immediately without retrying.
+`api_get_request/3` and `api_post_request/5` retry up to 5 times, sleeping between attempts for an equal-jitter exponential backoff interval: `Temp = min(10000, 500 * 2^Attempt)` ms, then a random delay in `[Temp/2, Temp]`. The delay is skipped after the final attempt, so a fully exhausted request waits at most 7.5s in total. Credentials are validated before the first attempt and re-validated before each retry. If credentials cannot be loaded, returns `{error, {credentials, Reason}}` immediately without retrying.
 
 ## Modules
 
--- a/include/aws_lib.hrl
+++ b/include/aws_lib.hrl
@@ -54,7 +54,8 @@
 % 5-minute buffer erlcloud uses.
 -define(CREDENTIAL_REFRESH_BUFFER_SECONDS, 300).
 
--define(LINEAR_BACK_OFF_MILLIS, 500).
+-define(BACKOFF_BASE_MILLIS, 500).
+-define(BACKOFF_CAP_MILLIS, 10000).
 -define(MAX_RETRIES, 5).
 
 -define(AWS_CREDENTIALS_TABLE, aws_credentials).
--- a/src/aws_lib.erl
+++ b/src/aws_lib.erl
@@ -28,6 +28,7 @@
     local_time/0,
     api_get_request/3,
     api_post_request/5,
+    backoff_delay/3,
     endpoint/4,
     sign_headers/10,
     instance_volumes/1,
@@ -582,7 +583,7 @@ api_get_request(Service, Path, State) ->
         "",
         [],
         ?MAX_RETRIES,
-        ?LINEAR_BACK_OFF_MILLIS,
+        ?BACKOFF_BASE_MILLIS,
         State
     ).
 
@@ -606,7 +607,7 @@ api_post_request(Service, Path, Body, Headers, State) ->
         Body,
         Headers,
         ?MAX_RETRIES,
-        ?LINEAR_BACK_OFF_MILLIS,
+        ?BACKOFF_BASE_MILLIS,
         State
     ).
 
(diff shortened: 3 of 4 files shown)


lukebakken (approved):
> [!NOTE]

Reviewed at 9f43c1f.

Replacing a fixed delay with equal-jitter exponential backoff is the right call here, and the implementation details hold up:

- `1 bsl Attempt` rather than `math:pow/2` keeps the arithmetic integral, so there is no float conversion and no `badarith` at large exponents.
- `max(1, Temp div 2)` keeps `rand:uniform/1` from ever seeing 0. I checked the degenerate inputs directly: `backoff_delay(0, 0, 0)`, `(0, 1, 1)` and `Cap < Base` all return positive values without crashing.
- Equal jitter over full jitter is well argued. A full-jitter draw can collapse to ~1ms, which defeats the point of backing off; the `Temp/2` floor still leaves a full `Temp/2` of spread to decorrelate concurrent retriers.
- Issue #80's non-retriable short-circuit is genuinely untouched, so a 4xx still returns immediately rather than sleeping through the retry budget.

Verification: eunit **821 passed, 0 failed**, against a **805** baseline measured on current `main` in a scratch worktree. Dialyzer reports 12 warnings, all in modules this PR does not touch (`aws_auth_validate_*`, `aws_lib_config.erl`, `aws_sup.erl`), none attributed to `aws_lib.erl`. `erlfmt` clean. All four `build-test` jobs and `format-check` are green.

I raised five findings on the original head, and all five are now resolved:

1. **The sleep after the final attempt was wasted.** The loop slept before recursing, so a fully exhausted request slept once per attempt instead of once per gap. Measured with a stubbed `timer`: 5 attempts but 5 sleeps, the last one ~7.75s that bought no further attempt. Under the old fixed 500ms delay that waste cost 500ms, so exponential backoff amplified it by roughly 16x, on the broker-boot ARN resolution path. Fixed in 9f43c1f by gating the sleep behind `maybe_backoff/3`; the same run now performs 5 attempts and 4 sleeps, and worst-case cumulative delay drops from 15.5s to 7.5s (expected ~11.6s to ~5.6s).
2. **The branch conflicted with `main` and would have reverted #170.** The conflict hunk was the export list: `main` had deleted `open_connection/2,3`, `close_connection/1` and `direct_request/7` while this branch added `backoff_delay/3` beside them, so resolving the wrong way would have silently restored 147 lines of dead code. Resolved by rebasing onto ed071a3, keeping the deletions.
3. `AGENTS-aws_lib.md` still described "a fixed 500ms delay between attempts", naming a policy the code no longer implements. Updated.
4. A comment cited `api_request_with_retries/10`; threading the attempt counter made the recursive clause arity 11. Corrected.
5. `backoff_delay/3` was added to the public export list, but `export_all` under `TEST` already gives the suite access, as it does for the unexported `api_request_with_retries/8`. Dropped, which also removes the line that caused the conflict in finding 2.

Full disclosure for @sdewhitt: 9f43c1f was written under @lukebakken's direction to address findings 1, 3, 4 and 5, and the description's verification and latency numbers were refreshed to match (the previous figures were pinned to a pre-#170 baseline). That commit therefore has not had an independent third reviewer, so it is worth a look on your side.

One item is a judgment call rather than a defect, and I have deliberately not treated it as blocking: whether 7.5s worst case is acceptable on the broker-boot path is a product decision, not something the code can settle. Lowering `?BACKOFF_CAP_MILLIS` is the lever if it is too long. Deferring the caller-configurable version of that knob to #85 is the right scoping.

The merge-order request still stands: this should land before #85, which is still open and will move the same function's arity again.

LGTM.

PR #179, opened 2026-08-05
Run dialyzer in CI and clear the existing warning backlog

### LINKED ISSUE #8: GH actions - dialyzer
Dialyzer should be run as part of GHA

lukebakken:
Closes #8.

## Summary

`make dialyze` had a backlog of warnings, so it exited non-zero and could not be wired into CI. This clears the backlog to zero and adds the CI step.

The backlog was **12 warnings, not the 30 the issue lists**. Earlier merges (including 684a329, which extracted the retry loop into `aws_lib_retry`) had already cleared the entire `aws_auth_validate_ldap` "will never be called" cluster that the issue singles out.

Of the 12: two were real defects, two were dead code, and four were false positives.

## The registry override was broken, and caused most of the backlog

`effective_allowed_fields/2` read its per-method override like this:

```erlang
application:get_env(aws, {auth_validation_allowed_fields_override, Method})
```

`application:get_env/2` requires an atom parameter name, so with a `{atom(), binary()}` tuple the call could never succeed. **The allowed-fields override never took effect.** It is re-keyed the way `auth_validation_enabled_methods` directly below it already does: a single atom key holding a `[{Method, Fields}]` list.

The cascade is why this accounts for most of the backlog. Because dialyzer treats that call as always-failing, `effective_allowed_fields/2` has no local return, so `dispatch/2` cannot return normally through its `{ok, Module}` branch and its success type collapses to `{error, method_disabled | unknown_method}`. That makes six downstream patterns in `aws_auth_validate_mgmt` look unreachable. Fixing the key cleared **five warnings beyond the two at the call site**.

The eunit setup set the old tuple key, so it moves to the new shape. Left alone it would have kept passing while silently no longer exercising the override.

## Two clauses were genuinely dead and are removed

- `ini_parse_line_parts(Section, [])` in `aws_lib_config`: `ini_split_line/1` always returns a non-empty list, and nothing else calls the function. Spec narrowed to `[string(), ...]`
- the `infinity` bound in `aws_sup:get_int_env/3`: the function is local with a single call site passing `100`, so the branch was speculative generality. The bound is folded into the guard

## Four are false positives

Each keeps its clause and gains a targeted `-dialyzer({no_match, F/A})` plus a comment recording why the clause stays.

- `status_for_category/1` is reachable only through the `-ifdef(TEST)` export, and the dialyze target compiles without `TEST`, so dialyzer sees a local function and infers its domain from the sole production call site in `respond/3`. This is exactly the case the issue asked to confirm before suppressing: reachable via a test-only export, not dead code. A `-spec` alone cannot silence it, because the inferred success typing of a local function wins, so the spec documents the real domain and the suppression does the rest
- `peer_ip/1` and `format_ip/1` are fail-closed audit guards. `peer_ip/1` runs *before* the `try` in `with_semaphore/6`, so a `function_clause` there would surface as an unhandled 500 on an otherwise valid request
- `check_peer_ip/1` is the one case where the library spec is wrong rather than the code. `eldap:info/1` is specced to return `connection_info()` unconditionally, but it returns `recv/1`'s result, which is `{error, {internal_error, Reason}}` if the eldap process exits first. The fallback is reachable in practice, and dropping it would turn a dead connection into a crash instead of `blocked`

## The CI step

Added to `.github/workflows/build-test.yaml`, gated on `matrix.rmq-version == 'v4.3.x'` rather than run across all four matrix entries. Dialyzer findings depend on the OTP version because library specs change between releases, so running it everywhere would let a warning appearing under only one OTP release fail builds unrelated to the change under review.

## Verification

| Gate | Result |
| --- | --- |
| `make dialyze` | exit 0, `done (passed successfully)` |
| `erlfmt -c` | clean |
| `make eunit` | 834 passed |
| `make ct-aws_auth_validate_mgmt` | 20 passed, 0 failed, 3 skipped |

The 3 skips are the pre-existing OAuth-authz skips (`OAuth authz evaluation unavailable on this broker node`), unrelated to this change.

Verified locally only under OTP 27.3.4.14, which is what the `v4.3.x` matrix entry uses. OTP 26 and 28 were not verified locally; CI covers those for build and tests, and by design does not run dialyzer on them.

## Reviewer note

Re-keying the override is a breaking change to a setting that never worked, so nothing deployed can depend on the old tuple key. The key is undocumented in `priv/schema/aws.schema` either way, so there is no schema change to make here. Flagging it explicitly in case you would rather see the override documented or removed outright instead.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -25,6 +25,7 @@ jobs:
           - rmq-version: 'v4.3.x'
             otp-version: 27
             elixir-version: 1.18
+            dialyze: true
           - rmq-version: 'v4.2.x'
             otp-version: 27
             elixir-version: 1.18
@@ -43,7 +44,8 @@ jobs:
         with:
           script: |
             core.setFailed('could not restore cached rabbitmq-server build');
-      - uses: erlef/setup-beam@54075bcc5e249e4758d363f27d099f55d843f124 # v1.24.1
+      - id: beam
+        uses: erlef/setup-beam@54075bcc5e249e4758d363f27d099f55d843f124 # v1.24.1
         with:
           otp-version: ${{ matrix.otp-version }}
           elixir-version: ${{ matrix.elixir-version }}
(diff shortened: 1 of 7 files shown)


.github/workflows/build-test.yaml
```diff
@@ -65,6 +65,13 @@ jobs:
           sudo aa-teardown || true
           sudo systemctl disable --now apparmor.service || true
       - run: make -C ${{ github.workspace }}/rabbitmq-server/deps/aws
+      # Dialyzer findings depend on the OTP version (library specs change
+      # between releases), so run it on a single configuration. Running it
+      # across the whole matrix would let a warning that only appears under one
+      # OTP release fail builds unrelated to the change under review.
+      - name: Run dialyzer
+        if: matrix.rmq-version == 'v4.3.x'
```
the-mikedavis: In some other projects I see `dialyze` being a boolean set in the matrix (defaulting to false something IIRC?). That might be easier to maintain in the long-run, so that all matrix parameters are in one place.

the-mikedavis (approved):

.github/workflows/build-test.yaml
```diff
@@ -65,6 +65,13 @@ jobs:
...
+        run: make -C ${{ github.workspace }}/rabbitmq-server/deps/aws dialyze
```
the-mikedavis, line 101: You might also want to cache the dialyzer artifacts since the PLT generation can take a while. I think we do this in `amazon-mq/rabbitmq-stream-s3` for example

sdewhitt (approved):

.github/workflows/build-test.yaml
lukebakken: Done in 24fcb25: the step now gates on a dialyze: true flag on the v4.3.x matrix entry, so the config lives in the matrix block.

.github/workflows/build-test.yaml
lukebakken: Done in 24fcb25. Keyed on runner OS, resolved OTP version (setup-beam output, so exact patch), rmq-version, elixir-version, and a Makefile hash (no rebar.lock/mix.lock in this project). No restore-keys on purpose: make dialyze only rebuilds an existing PLT on an erts-path change, not a dep-set change, so a near-miss restore would analyze against a stale PLT. A key change is a clean miss that rebuilds and re-caches.

PR #181, opened 2026-08-05
Make auth_validation allowed_fields override operator-configurable

### LINKED ISSUE #180: auth_validation_allowed_fields_override is unreachable for operators: finish or remove it
## Summary

`auth_validation_allowed_fields_override` is unreachable for operators. It should either be finished or removed. Splitting this out of #179, where fixing the key was necessary to clear a dialyzer cascade but deliberately did not decide the feature's fate.

## Nothing can set it

There are exactly three references in the repo:

- the `application:get_env/2` read in `aws_auth_validate_registry:effective_allowed_fields/2`
- a `set_env` / `unset_env` pair inside `registry_field_filter_override_test_/0`

Beyond those:

- there is **no mapping in `priv/schema/aws.schema`**, so no `rabbitmq.conf` key reaches it, and an operator has no supported way to configure it
- `PROJECT_ENV` in the `Makefile` is `[]`, so there is no application-env default either
- nothing in `src/` calls `set_env` for it

In production the `get_env` therefore always returns `undefined`, control falls to the `_ ->` branch, and `effective_allowed_fields/2` returns `BaseFields` unchanged. The only caller is `dispatch/2`, which passes the result to `maps:with/2`.

## Its only exercise is the test that tests it

`registry_field_filter_override_test_/0` sets the key itself and then calls `effective_allowed_fields/2` directly. It is a unit test of a code path no other code can reach.

## It never worked, even in principle, before #179

The override was introduced in e94e192 (the commit that added the registry) reading a **tuple** key:

```erlang
application:get_env(aws, {auth_validation_allowed_fields_override, Method})
```

`application:get_env/2` requires an atom parameter name, so that call always broke its contract and the override branch was unreachable regardless of configuration. It has been non-functional since it landed.

#179 re-keys it to a single atom key holding a `[{Method, Fields}]` list, matching how `auth_validation_enabled_methods` is read directly below it. That makes the key **legal**, which is what cleared the dialyzer cascade (the always-failing call gave `effective_allowed_fields/2` no local return, collapsing `dispatch/2`'s success type and making six downstream patterns in `aws_auth_validate_mgmt` look unreachable). It does **not** make the feature reachable by an operator. The state goes from "impossible" to "possible only if something sets an undocumented app env".

## Options

1. **Remove it.** Drop the override branch from `effective_allowed_fields/2`, drop its test, and inline `Module:allowed_fields()` in `dispatch/2`. Simplest, and deletes the construct that caused half the warning backlog in #8
2. **Finish it.** Add an `aws.auth_validation.allowed_fields.$method` mapping plus translation, wired the same way `aws.auth_validation.enabled_methods.$name` already is, so it becomes a real operator-facing control. Also document it in `README.md`
3. **Leave it.** Keep it as a legal but undocumented escape hatch

## Recommendation

Option 1, with a low degree of confidence, because this is a product call rather than a technical one.

The reasoning: the override can only ever **narrow** the backend's own `allowed_fields()` list, never widen it, since every entry is intersected with `BaseFields`. So it is defense-in-depth on top of a filter that already exists and already defaults to the safe set. That is a thin benefit against carrying an undocumented, untriggerable setting.

The counter-argument, and the reason for assigning @sdewhitt: it may have been deliberate scaffolding for something planned, in which case option 2 is the better outcome and the schema work is small.

## Note on option 2, if chosen

A `$method` mapping would need to reject unknown method names at translation time, or an operator typo would silently produce an override that never matches, which is the same class of failure this issue is about.

---

*This issue description was drafted by an AI agent and reviewed by @lukebakken.*

sdewhitt:
Resolves #180 with option 2 (finish it): makes the per-method field allowlist a real operator-facing control.

Lands after #179 (`fix/dialyzer-warnings-and-ci`)

## Changes
- **`priv/schema/aws.schema`**: add `aws.auth_validation.allowed_fields.$method` mapping + translation, emitting the `[{Method, Fields}]` list into `auth_validation_allowed_fields_override` (mirrors `enabled_methods`). Unknown method names are rejected at config-generation time via `cuttlefish:invalid/2`, so a typo fails the boot instead of silently producing a never-matching override.
- **Tests**: extend the override unit test (add an un-listed-method case) and add a `config_schema` snippet.
- **`AUTH_VALIDATION.md`**: document the new key.

Narrow-only intersection semantics are unchanged: the override can only remove fields, never widen.

## Verification
- `ct-config_schema`: pass
- `eunit` (`aws_auth_validate_tests`): 183/183 pass
- Unknown-method rejection confirmed against the compiled schema
- `erlfmt -c`: clean

### DIFF
--- a/AUTH_VALIDATION.md
+++ b/AUTH_VALIDATION.md
@@ -148,6 +148,30 @@ and no log entry. Setting `max_body_size = 2000000`, for example, leaves the
 limit at `65536`, and a 200 KB body is then rejected as `body_too_large` with
 nothing pointing at the ignored setting.
 
+### Narrowing accepted request fields (optional)
+
+Each backend already accepts only the fields relevant to it (its
+`allowed_fields`); any other top-level field is silently dropped. If you want to
+narrow that set further -- for example to reject `queries` and `dn_lookup_base`
+on the `ldap` method and accept only the basic bind fields -- set a per-method
+allowlist:
+
+```
+aws.auth_validation.allowed_fields.ldap = servers,port,user_dn,password_arn,use_ssl
+```
+
+The value is a comma-separated list of field names. The effective set is the
+**intersection** of this list and the backend's own defaults, so the key can
+only ever remove fields, never add one; an entry that is not a real field for
+that method simply never matches. A method left unset keeps the backend's full
+default set. Unknown method names (anything other than `ldap`, `http`, `oauth`,
+`tls`) are rejected at configuration-generation time rather than silently
+ignored, so a typo fails the boot rather than producing an allowlist that never
+applies.
+
+This is a defense-in-depth control on an endpoint that is already admin-gated,
+opt-in, and side-effect-free; most deployments will not need it.
+
 The plugin also registers an **Auth Validation** tab in the RabbitMQ management
 console UI that drives the same endpoint. Note that the tab is registered
 whenever the plugin itself is enabled, independently of the toggles above: with
--- a/priv/schema/aws.schema
+++ b/priv/schema/aws.schema
@@ -129,6 +129,42 @@
      [{list_to_binary(lists:last(Key)), Value} || {Key, Value} <- Settings]
  end}.
 
+%% @doc Optional per-method request-field allowlist. Narrows the fields a
+%% backend will accept to the intersection of this list and the backend's own
+%% allowed_fields/0; it can only remove fields, never add one, so an operator
+%% cannot use it to widen the surface. Leave a method unset to accept the
+%% backend's full default set. The value is a comma-separated list of field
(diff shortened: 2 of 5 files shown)


the-mikedavis (approved):
Looks good to me to merge after #179

lukebakken (approved):
Approving. Cleanly promotes the previously-unreachable allowed_fields override (issue #180) into an operator control: schema mapping and translation mirror the existing enabled_methods pattern, unknown methods are rejected at config-generation time, and the runtime intersection can only ever narrow the backend's own allowed_fields, never widen them.

Verified locally: compile, dialyzer, eunit (183 tests), and the config_schema run_snippets suite all pass.

PR #183, opened 2026-08-06
Add optional credentialed-probe mode to HTTP auth validation

### LINKED ISSUE #182: Auth Validation: add an optional credentialed-probe mode to `PUT /api/aws/auth/validate/http`
## Summary

The `http` validation backend (`aws_auth_validate_http`) is currently **reachability-only**: it confirms each configured `*_path` is reachable over (m)TLS and answers with an auth-server-shaped `allow`/`deny` body, using a **synthetic** probe principal (`__rabbitmq_config_validation_probe__`). It deliberately does **not** assert that any particular principal is authorized -- a `deny` for the synthetic user is treated as success, because a well-formed `deny` still proves the endpoint speaks the auth protocol.

This issue tracks the follow-on **credentialed-probe mode** already flagged as out of scope in the initial cut (`src/aws_auth_validate_http.erl:42`):

> A future credentialed-probe mode (assert `user_path` returns `allow` for a supplied username + `password_arn`) is intentionally out of scope for this initial reachability-only implementation.

## Motivation

Reachability-only catches the common pain (wrong URL, TLS/CA misconfig, unreachable host, path pointing at a non-auth service) without needing a real test credential. It cannot catch a server that is reachable and auth-shaped but that would **reject a credential the operator expects to succeed** -- e.g. a backend pointed at the wrong user store, a misconfigured shared secret, or a `user_path` that returns `deny` for every real user. An operator validating a config before a restart currently has no way to confirm a known-good credential
actually authenticates.

## Proposed behavior

When (and only when) the request supplies a credential, the probe against `user_path` uses that credential instead of the synthetic principal and asserts the response is `allow` (not merely auth-shaped).

Proposed new `allowed_fields` for the `http` method:
- `username` (inline, non-secret identifier)
- `password_arn` (secret resolved server-side via Secrets Manager, mirroring the LDAP backend -- never passed inline)

Behavior:
- **No `username`/`password_arn` supplied** -> unchanged reachability-only path (synthetic principal, `allow`-or-`deny` both pass). Fully backward compatible.
- **`username` + `password_arn` supplied** -> resolve `password_arn`, probe `user_path` with the real params, and:
  - `allow` -> success (204)
  - well-formed `deny` -> **`auth_failed` (422)** (the credential the operator expected to work did not) -- distinct from today, where a synthetic `deny` is a pass
  - not auth-shaped / unreachable / TLS failure -> existing categories (`auth_failed` shape mismatch, `connection_failed`, `tls_failed`)
- **`password_arn` without `username`, or vice versa** -> `input_invalid` (400) (both required together).
- The other paths (`vhost_path`, `resource_path`, `topic_path`) remain reachability-only with the synthetic principal; credentialed mode is `user_path`-only, matching how `rabbit_auth_backend_http` issues the user check.

## Constraints / invariants to preserve

- **ARN-first ordering**: resolve `password_arn` only after all pure input validation passes (same as the LDAP backend's `resolve_request_state/1`), so a malformed request never triggers a Secrets Manager fetch.
- **assume_role guardrail**: resolving `password_arn` means the request now resolves an ARN, so a configured `aws.arns.assume_role_arn` becomes MANDATORY for credentialed mode -- else `config_conflict` (422), matching the LDAP backend. (Today the HTTP backend only requires assume_role when TLS-material ARNs are referenced.)
- **No credential leakage (R6)**: the resolved password must never appear in logs, responses, or crash reports. It is a probe parameter only.
- **Parity**: `params_for/1` currently builds probe params fed to both the probe and the parity test; the credentialed path must remain parity-checked against what `rabbit_auth_backend_http` sends for a real user check (`do_http_req/2`).
- **Fixed category set**: prefer reusing `auth_failed` (422) for "expected credential was denied". Only introduce a dedicated category (e.g. `endpoint_unverified`, mirroring `authz_unverified`) if `auth_failed` proves too coarse -- and if so, extend both `aws_auth_validate_backend:error_category/0` and the handler's `status_for_category/1`.

## Scope

- [ ] Add `username` + `password_arn` to `aws_auth_validate_http:allowed_fields/0`
- [ ] Input validation: both-or-neither, non-empty, `password_arn` is a valid ARN
- [ ] Make `assume_role_arn` mandatory when `password_arn` is present
- [ ] Credentialed `user_path` probe: real params, `allow`-required semantics
- [ ] Parity test: credentialed probe params match `rabbit_auth_backend_http`
- [ ] Unit tests: allow -> 204; deny -> 422; missing-half -> 400; no-assume_role
      -> 422; password never logged
- [ ] Update `AUTH_VALIDATION.md` (http method description + per-method fields)
- [ ] Sync the console FE contract (new optional fields) in
      `AmazonMqConsoleAssets-React`

## Out of scope

- Credentialed probes for `vhost_path` / `resource_path` / `topic_path`
  (authz-semantics judgement, not a reachability/credential check).
- Any change to the synthetic-principal reachability path when no credential is
  supplied.

## References

- `src/aws_auth_validate_http.erl:29-44` (reachability-only scope note)
- `src/aws_auth_validate_http.erl:680-691` (`query_for/1` / `params_for/1`,
  `?PROBE_USER`)
- `src/aws_auth_validate_ldap.erl` (`password_arn` + assume_role guardrail to mirror)

sdewhitt:
_Closes #182_

Adds an optional credentialed-probe mode to `PUT /api/aws/auth/validate/http`.

Today the `http` method is reachability-only: it confirms each configured `*_path` is reachable and answers with an auth-shaped `allow`/`deny` body, using a synthetic principal. This adds an opt-in mode that probes `user_path` with a real credential and asserts the response is `allow`.

- New optional fields `username` (inline) and `password_arn` (resolved server-side; never passed inline). Supplying one without the other is rejected as `input_invalid`.
- With no credential supplied, behaviour is unchanged (fully backward compatible).
- Because credentialed mode resolves an ARN, a configured assume-role becomes required, else `config_conflict`.
- A well-formed `deny` for the supplied credential returns `auth_failed`; the resolved password never appears in logs, responses, or crash reports.

### Tests
`allow` -> 204, `deny` -> 422, missing-half -> 400, ARN without assume-role -> 422, and query-encoding of special characters. Full suite green.

### DIFF
--- a/src/aws_auth_validate_http.erl
+++ b/src/aws_auth_validate_http.erl
@@ -296,46 +296,25 @@ validate(Body) when is_map(Body) ->
 %% into Params. Runs after all pure input validation has passed, so a malformed
 %% request never triggers an STS AssumeRole call (ARN-first ordering).
 %%
-%% Mirrors the LDAP backend's guardrail (resolve_request_state/1 there): when the
-%% request resolves ANY ARN, a configured `aws.arns.assume_role_arn' is
-%% MANDATORY. We assume that role -- the SAME role the plugin already assumes at
-%% boot to resolve every configured ARN (aws_arn_config:maybe_assume_role/1) --
-%% into a request-local aws_lib state and resolve the TLS-material ARNs under it.
-%% This is operator config, not caller input, so it raises no confused-deputy
-%% concern.
-%%
-%% When NO role is configured we do NOT fall back to a default aws_lib state:
-%% that would resolve ARNs with the broker's ambient (EC2 instance) credentials,
-%% which on Amazon MQ can be far more privileged than the role a customer would
-%% attach to their own secret/bucket, so a validate request could resolve ARNs
-%% the caller's intended role never could -- a least-privilege pitfall. We never
-%% use the instance role: with an ARN referenced and no assume_role configured,
-%% refuse with config_conflict before any secret fetch or outbound connection.
-%%
-%% For the credentialed mode (username + password_arn): the password_arn will be
+%% Delegates to the shared aws_auth_validate_ssl:resolve_request_state/2 for both
+%% modes. In credentialed mode (username + password_arn) the password_arn will be
 %% resolved under the assumed role, so an assume_role_arn is MANDATORY regardless
-%% of whether TLS material ARNs are also referenced. This matches the LDAP
-%% backend where password_arn always forces the role. For reachability-only mode,
-%% the behaviour is unchanged: ARNs are only required when TLS material is
-%% referenced.
+%% of whether TLS material ARNs are also referenced -- the `force_assume_role'
+%% flag in the opts map achieves this by skipping the request_references_arn check
+%% and always requiring the configured role.
 %%
-%% Delegates to the shared aws_auth_validate_ssl helper for reachability mode.
-%% The no-ARN branch there stores the `none' credential sentinel (which
-%% resolve_arn/2 refuses), matching the fail-closed contract PR #116 introduced
-%% -- a customer request can never resolve an ARN under the broker's ambient EC2
-%% instance role.
+%% For reachability-only mode the behaviour is unchanged: the shared helper checks
+%% whether any ARN key is present in ssl_options before requiring a role. The
+%% no-ARN branch stores the `none' credential sentinel (which resolve_arn/2
+%% refuses), matching the fail-closed contract PR #116 introduced -- a customer
+%% request can never resolve an ARN under the broker's ambient EC2 instance role.
 resolve_request_state(#{credential_mode := credentialed} = Params) ->
     %% Credentialed mode resolves password_arn, so assume_role is mandatory
-    %% regardless of whether TLS material ARNs are also referenced.
-    case aws_auth_validate_ssl:configured_assume_role_arn() of
-        none ->
-            {error, config_conflict, ?REASON_NO_ASSUME_ROLE};
-        RoleArn ->
-            case aws_iam:assume_role(RoleArn, aws_lib:new()) of
-                {ok, State} -> {ok, Params#{aws_state => State}};
-                {error, _} -> {error, input_invalid, ?REASON_ASSUME_ROLE}
-            end
-    end;
+    %% regardless of whether TLS material ARNs are also referenced. The shared
+    %% helper's force_assume_role flag skips the request_references_arn check and
+    %% always requires/assumes the configured role.
+    Opts = (ssl_opts())#{force_assume_role => true},
+    aws_auth_validate_ssl:resolve_request_state(Params, Opts);
 resolve_request_state(Params) ->
     aws_auth_validate_ssl:resolve_request_state(Params, ssl_opts()).
 
@@ -600,7 +579,18 @@ probe_one(Key, Url, Params, SslOpts, Profile) ->
         {ok, PinnedUrl, Host} ->
             UrlStr = maps:get(url_string, PinnedUrl),
             {Query, Classifier} = probe_query_and_classifier(Key, Params),
-            Request = build_request(Method, UrlStr, Query, Host),
+            %% Force POST for credentialed probes: a GET would append the
+            %% resolved password to the URL query string, leaking it into auth
+            %% server access logs and any intermediate proxy/LB logs. POST puts
+            %% credentials in the request body where they are not logged by
+            %% standard infrastructure. Reachability probes use whatever method
+            %% the operator configured (default GET) -- they carry no secret.
+            EffectiveMethod =
+                case maps:get(credential_mode, Params, reachability) of
+                    credentialed -> post;
+                    _ -> Method
+                end,
+            Request = build_request(EffectiveMethod, UrlStr, Query, Host),
             HttpOpts =
                 [
                     {timeout, Timeout},
@@ -610,7 +600,9 @@ probe_one(Key, Url, Params, SslOpts, Profile) ->
                     %% do not re-vet the Location target).
                     {autoredirect, false}
                 ] ++ ssl_http_opt(Url, Host, SslOpts),
-            case httpc:request(Method, Request, HttpOpts, [{body_format, binary}], Profile) of
+            case
+                httpc:request(EffectiveMethod, Request, HttpOpts, [{body_format, binary}], Profile)
+            of
                 {ok, {{_Vsn, Code, _Phrase}, _Headers, Body}} ->
                     case lists:member(Code, ?AUTH_RESPONSE_CODES) of
                         %% A usable status is necessary but not sufficient: the
@@ -641,11 +633,13 @@ probe_query_and_classifier(Key, _Params) ->
 
 %% Build the query string for a credentialed user_path probe, matching the
 %% fields rabbit_auth_backend_http sends for a user check (username + password).
+%% Uses rabbit_http_util:quote_plus/1 -- the same encoder the real backend uses
+%% in rabbit_auth_backend_http:q/1 -- to ensure byte-for-byte encoding parity.
+%% uri_string:compose_query would re-encode non-ASCII bytes as UTF-8 codepoints
+%% (double-encoding), producing different percent-sequences for multi-byte chars.
 credentialed_query_for(#{username := Username, password := Password}) ->
-    uri_string:compose_query([
-        {"username", binary_to_list(Username)},
-        {"password", binary_to_list(Password)}
-    ]).
+    "username=" ++ rabbit_http_util:quote_plus(Username) ++
+        "&password=" ++ rabbit_http_util:quote_plus(Password).
 
 %% Map an httpc transport error to a fixed category (shared classifier). A
 %% TLS/cert failure -> tls_failed; everything else -> connection_failed. The raw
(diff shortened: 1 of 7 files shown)


src/aws_auth_validate_http.erl
```diff
@@ -552,14 +618,35 @@ probe_one(Key, Url, #{http_method := Method, timeout := Timeout}, SslOpts, Profi
...
+    {Query, Classifier};
+probe_query_and_classifier(Key, _Params) ->
+    Query = query_for(Key),
+    Classifier = fun(Body) -> classify_response(Key, Body) end,
+    {Query, Classifier}.
+
+%% Build the query string for a credentialed user_path probe, matching the
+%% fields rabbit_auth_backend_http sends for a user check (username + password).
+credentialed_query_for(#{username := Username, password := Password}) ->
+    uri_string:compose_query([
```
lukebakken: **Non-ASCII password/username is double-encoded, producing a false `auth_failed`.** This uses `uri_string:compose_query`, but the backend this probe claims to mirror (`rabbit_auth_backend_http:q/1`) uses `rabbit_http_util:quote_plus`, which percent-encodes each raw byte. `compose_query` instead treats each byte of `binary_to_list/1` as a Unicode codepoint and re-UTF-8-encodes it. Reproduced at runtime for a `café` password:

```
password=caf%C3%A9        <- real backend (quote_plus)
password=caf%C3%83%C2%A9  <- this PR (compose_query)
```

Any password containing a non-ASCII byte (and likewise a non-ASCII `username`) is corrupted on the wire: the auth server decodes a different password, returns `deny`, and the probe reports 422 `auth_failed` for a credential the live broker would accept. That defeats the module's stated fidelity guarantee. The existing parity test (`aws_auth_validate_http_parity_tests.erl`) reasoned this divergence "diverge-but-safe", but only covers ASCII special chars, where it happens to hold; it does not for the multi-byte values credentialed mode newly sends. Suggest encoding with `rabbit_http_util:quote_plus` (as the real backend does) so bytes match, and extending the parity test with a non-ASCII value.

src/aws_auth_validate_http.erl
```diff
@@ -552,14 +618,35 @@ probe_one(Key, Url, #{http_method := Method, timeout := Timeout}, SslOpts, Profi
...
+    Classifier = fun classify_credentialed_response/1,
+    {Query, Classifier};
+probe_query_and_classifier(Key, _Params) ->
+    Query = query_for(Key),
+    Classifier = fun(Body) -> classify_response(Key, Body) end,
+    {Query, Classifier}.
+
+%% Build the query string for a credentialed user_path probe, matching the
+%% fields rabbit_auth_backend_http sends for a user check (username + password).
+credentialed_query_for(#{username := Username, password := Password}) ->
```
lukebakken, line 640: **Default GET method places the real resolved password in the URL query string.** In credentialed mode with `http_method` omitted (defaults to `get` at line 451) or `get`, this `username=..&password=<secret>` query is appended to the URL by `build_request/4`'s get clause. That URL then lands in the target auth server's HTTP access logs and any proxy/LB in between -- surfaces the broker does not control, which the documented "password is never logged" invariant does not cover. The pre-PR reachability probe only ever put a synthetic username in the URL. Forcing POST for the credentialed probe keeps the secret out of the access-log surface (a POST body is not logged there).

Related, lower-confidence: the password is embedded in the httpc `Request` tuple handed to inets (the `httpc:request/5` call in `probe_one/5`). inets services the request in its own `httpc_handler` process, which holds the `#request{}` (URI + body) in gen_server state and runs OUTSIDE this module's `do_http_validate/1` try/catch and the handler's `with_semaphore/6` catch. A crash in that inets process could emit an OTP crash_report carrying the password. POST does not fully close this, but it removes the access-log exposure, which is the concrete one.

src/aws_auth_validate_http.erl
```diff
@@ -600,6 +687,17 @@ classify_response(Key, Body) ->
         false -> {error, auth_failed, ?REASON_ENDPOINT}
     end.
 
+%% In credentialed mode, only `allow' is success. A well-formed `deny' means
+%% the credential was rejected -- the operator expected it to work, so this is
+%% auth_failed. A non-auth-shaped body is also auth_failed (same as a deny: the
+%% server did not grant access for the supplied credentials).
+classify_credentialed_response(Body) ->
```
lukebakken, line 691: **A reachable non-auth endpoint is misreported as a credential denial.** The reachability classifier distinguishes "endpoint is not an auth server" (`classify_response` -> `?REASON_ENDPOINT`) from a well-formed `deny`. This credentialed classifier collapses both: any body that is not an `allow` prefix returns `auth_failed` / "HTTP auth server denied the supplied credentials". Failure scenario: `user_path` (with a valid username + password_arn) points at a health check or proxy returning 200 `OK`. The probe reports a credential denial, steering the operator to debug IAM/passwords when the real fault is a wrong URL pointing at a non-auth service. Consider keeping the `?REASON_ENDPOINT` (not-an-auth-server) distinction for a body matching neither allow nor deny, and reserving `?REASON_AUTH_DENIED` for a well-formed `deny`.

test/aws_auth_validate_http_credentialed_tests.erl
```diff
@@ -0,0 +1,154 @@
+%% Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
lukebakken, line 1: **No integration coverage of the credentialed probe path.** `aws_auth_validate_http_SUITE` drives `validate/1` through an in-process httpd stub for every reachability outcome (allow/deny/404/500/TLS/mTLS) -- that is the customary bar here. The credentialed path (`probe_query_and_classifier` -> `credentialed_query_for` -> real query -> `classify_credentialed_response`) has no coverage at that level; these new tests only exercise the classifier/parser as isolated functions. A wiring bug (wrong classifier on a path, or the credential query not actually threaded into the request) would pass every unit test yet break the real probe -- which is exactly how findings 1 and 3 slip through. Suggest a SUITE case that stubs an `allow` and a `deny` on user_path and asserts `ok` vs `auth_failed` end to end.

src/aws_auth_validate_http.erl
```diff
@@ -303,25 +312,58 @@ validate(Body) when is_map(Body) ->
...
+%% backend where password_arn always forces the role. For reachability-only mode,
+%% the behaviour is unchanged: ARNs are only required when TLS material is
+%% referenced.
+%%
+%% Delegates to the shared aws_auth_validate_ssl helper for reachability mode.
+%% The no-ARN branch there stores the `none' credential sentinel (which
+%% resolve_arn/2 refuses), matching the fail-closed contract PR #116 introduced
+%% -- a customer request can never resolve an ARN under the broker's ambient EC2
+%% instance role.
+resolve_request_state(#{credential_mode := credentialed} = Params) ->
```
lukebakken, line 311: **The credentialed-mode security guardrail is untested.** Nothing asserts the two new branches: credentialed mode with no `assume_role_arn` -> `config_conflict` (AUTH_VALIDATION.md documents this as mandatory), and assume-role failure -> `input_invalid`. This clause is the guardrail that forces `password_arn` to resolve under an operator role rather than the broker instance role, so a regression that let credentialed mode proceed without a configured role would not be caught. Note `password_never_in_error_test` only checks `arn_resolve_reason/1` in isolation, not the real `validate/1` / `resolve_credential/1` path, so there is also no end-to-end assertion that the resolved HTTP password never appears in a response.

lukebakken, line 311: Altitude (lower priority): this clause re-implements the `configured_assume_role_arn` + `aws_iam:assume_role` + `config_conflict`/`input_invalid` mapping that already lives in `aws_auth_validate_ssl:resolve_request_state/2`. That shared helper's own header says it exists specifically to stop this role logic from being duplicated and drifting (a security fix would otherwise have to be applied in multiple places). The credentialed case is just "force role required regardless of TLS ARNs", which could be expressed by extending the shared helper (an opts flag, or adding `password_arn` to its `arn_keys`) rather than a bespoke duplicate that also re-references `?REASON_NO_ASSUME_ROLE` / `?REASON_ASSUME_ROLE` outside the `ssl_opts()` reasons map.

lukebakken (changes requested):
Thanks for adding the credentialed-probe mode. The mode dispatch (probe_query_and_classifier keeping probe_one generic) is clean, the ARN-first ordering is preserved, and the assume-role-mandatory guardrail for password_arn is the right call. A few issues need addressing before merge: findings 1-3 are the blockers (a functional encoding bug, a secret-exposure surface, and a misleading diagnostic), 4-5 are the coverage gaps that let 1 and 3 slip through, and 6 is maintainability.

Verified locally at deps/aws on issue-182 (HEAD d3a8a91): eunit passes (108), and I reproduced the encoding divergence in finding 1 against rabbit_http_util:quote_plus at runtime.

src/aws_auth_validate_http.erl
sdewhitt: Fixed in 74d1a26 -- `credentialed_query_for/1` now uses `rabbit_http_util:quote_plus`, byte-for-byte matching `rabbit_auth_backend_http:q/1`. Added a non-ASCII parity test asserting identical output to the real backend encoder.

src/aws_auth_validate_http.erl
sdewhitt: Fixed in 74d1a26 -- the credentialed probe now forces POST regardless of the configured/omitted `http_method`, keeping the password in the request body (out of the access-log surface). The reachability probe is unchanged.

src/aws_auth_validate_http.erl
sdewhitt: Fixed in 74d1a26 -- `classify_credentialed_response/1` now distinguishes a well-formed `deny` (`?REASON_AUTH_DENIED`) from a body matching neither allow nor deny (`?REASON_ENDPOINT`), reusing the reachability grammar. A non-auth 200 no longer reports a credential denial.

test/aws_auth_validate_http_credentialed_tests.erl
sdewhitt: Added in 74d1a26 -- new `aws_auth_validate_http_SUITE` cases drive `validate/1` end to end through the credentialed path: `allow`->ok, `deny`->auth_failed, and a non-auth 200 body->endpoint (not auth_failed).

src/aws_auth_validate_http.erl
sdewhitt: Added in 74d1a26 -- SUITE cases assert credentialed mode with no `assume_role_arn`->`config_conflict` and assume-role failure->`input_invalid`, plus an end-to-end assertion that the resolved password never appears in the error.

sdewhitt: Addressed in 74d1a26 -- rather than duplicate the logic, extended `aws_auth_validate_ssl:resolve_request_state/2` with an optional `force_assume_role` opt (defaults false, so LDAP/TLS/http callers are unaffected). The credentialed clause just sets that flag. (Follow-up 0ced90b declares the new key in the `opts()` type for dialyzer.)

the-mikedavis (approved):

PR #184, opened 2026-08-06
Add Prometheus metrics for auth validation

sdewhitt:
Adds Prometheus metrics for the auth-validation endpoint so operators can observe call volume, outcome mix, latency, and capacity pressure.

Metrics are emitted at the single point every request passes through (the audit log call site), dimensioned only by method and the fixed result-category set. No source IP, user, target, or secret material is ever used as a label.

Emitted via the `prometheus_collector` behaviour (the same pull-based pattern other RabbitMQ plugins use): the hot path is a lock-free counter update, and formatting happens only on scrape.

- `..._requests_total` (counter), `..._duration_ms` (histogram), `..._capacity_exhausted_total` (counter)
- semaphore in-use / capacity gauges
- Metrics register only when the feature is enabled; no state exists when it is off.

Adds new dependency, `prometheus`, to DEPS. However, it is already listed as a dependency in rabbitmq_prometheus, [which already exists within rabbitmq-server.](https://github.com/sdewhitt/rabbitmq-server/blob/aws/ldap-integration/deps/rabbitmq_prometheus/Makefile)

### Tests
CT suite covering counter increments per outcome, duration recording, capacity signalling, gauge behaviour, disabled-state, and crash-safety. All green.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ define PROJECT_ENV
 []
 endef
 
-DEPS = rabbit_common rabbit rabbitmq_management gun jose
+DEPS = rabbit_common rabbit rabbitmq_management prometheus gun jose
 TEST_DEPS = meck proper rabbitmq_ct_helpers rabbitmq_ct_client_helpers rabbitmq_auth_backend_ldap rabbitmq_auth_backend_http rabbitmq_auth_backend_oauth2
 LOCAL_DEPS = crypto inets ssl xmerl public_key eldap
 
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ define PROJECT_ENV
 []
 endef
 
-DEPS = rabbit_common rabbit rabbitmq_management prometheus gun jose
+DEPS = rabbit_common rabbit rabbitmq_management rabbitmq_prometheus gun jose
 TEST_DEPS = meck proper rabbitmq_ct_helpers rabbitmq_ct_client_helpers rabbitmq_auth_backend_ldap rabbitmq_auth_backend_http rabbitmq_auth_backend_oauth2
 LOCAL_DEPS = crypto inets ssl xmerl public_key eldap
 
--- a/src/aws_auth_validate_metrics.erl
+++ b/src/aws_auth_validate_metrics.erl
@@ -11,8 +11,14 @@
 %% just increments lock-free OTP `counters` stored in `persistent_term`.
 %%
 %% Counter layout (single counters ref, write_concurrency):
-%%   Slots 1..NUM_REQUESTS_SLOTS       -- requests_total per (method, category)
-%%   Slots HIST_BASE+1..HIST_BASE+NUM  -- histogram per method (buckets+sum+count)
+%%   Slots 1..75 (NUM_REQUESTS_SLOTS)  -- requests_total per (method, category)
+%%   Slots 76..140 (HIST_BASE+1..end)  -- histogram per method
+%%
+%% Per-method histogram (13 slots each):
+%%   Positions 1..10  -- finite bucket counters (le=10..10000)
+%%   Position 11      -- overflow bucket (le="+Inf", durations > 10000ms)
+%%   Position 12      -- sum of observed durations (NUM_BUCKETS+1)
+%%   Position 13      -- count of observations (NUM_BUCKETS+2)
 %%
 %% Semaphore gauges are read live from the semaphore gen_server at scrape time,
 %% not stored in counters.
@@ -24,7 +30,9 @@
 -export([deregister_cleanup/1, collect_mf/2]).
 
 -ifdef(TEST).
--export([counter_ref/0, method_index/1, category_index/1, histogram_base/0]).
+-export([
+    counter_ref/0, method_index/1, category_index/1, histogram_base/0, build_cumulative_buckets/2
+]).
 -endif.
 
 -import(prometheus_model_helpers, [create_mf/4]).
@@ -37,15 +45,19 @@
 
 %% Histogram bucket boundaries (milliseconds). Chosen to cover the range from
 %% fast local validation (~10ms) to slow DNS/TLS negotiation (~10s timeout).
+%% An 11th overflow bucket (slot 11) catches durations > 10000ms; emitted as
+%% le="+Inf" during scrape.
 -define(HISTOGRAM_BUCKETS, [10, 25, 50, 100, 250, 500, 1000, 2500, 5000, 10000]).
--define(NUM_BUCKETS, 10).
+-define(NUM_BUCKETS, 11).
 
-%% Per-method histogram layout: 10 bucket counters + 1 sum + 1 count = 12 slots.
--define(SLOTS_PER_HIST, 12).
+%% Per-method histogram layout: 11 bucket counters + 1 sum + 1 count = 13 slots.
+-define(SLOTS_PER_HIST, 13).
 
-%% Methods -- indexed 0..3 for counter math.
--define(METHODS, [<<"ldap">>, <<"http">>, <<"oauth">>, <<"tls">>]).
--define(NUM_METHODS, 4).
+%% Methods -- indexed 0..4 for counter math. The synthetic "unknown" method
+%% (index 4) collects requests_total for unrecognized method paths without
+%% exposing the raw caller-supplied string as a label (fixed cardinality).
+-define(METHODS, [<<"ldap">>, <<"http">>, <<"oauth">>, <<"tls">>, <<"unknown">>]).
+-define(NUM_METHODS, 5).
 
 %% Result categories -- indexed 0..14. This is the full set of categories
 %% that can appear in the audit trail (see aws_auth_validate_mgmt).
@@ -69,9 +81,9 @@
 -define(NUM_CATEGORIES, 15).
 
 %% Total counter slots:
-%%   requests_total: NUM_METHODS * NUM_CATEGORIES = 4 * 15 = 60
-%%   histogram: NUM_METHODS * SLOTS_PER_HIST = 4 * 12 = 48
-%%   Grand total: 108
+%%   requests_total: NUM_METHODS * NUM_CATEGORIES = 5 * 15 = 75
+%%   histogram: NUM_METHODS * SLOTS_PER_HIST = 5 * 13 = 65
+%%   Grand total: 140
 -define(NUM_REQUESTS_SLOTS, (?NUM_METHODS * ?NUM_CATEGORIES)).
 -define(HIST_BASE, ?NUM_REQUESTS_SLOTS).
 -define(TOTAL_SLOTS, (?NUM_REQUESTS_SLOTS + ?NUM_METHODS * ?SLOTS_PER_HIST)).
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -8,7 +8,7 @@ define PROJECT_ENV
 []
 endef
 
-DEPS = rabbit_common rabbit rabbitmq_management gun jose
+DEPS = rabbit_common rabbit rabbitmq_management prometheus gun jose
```
the-mikedavis: In rabbitmq-stream-s3 we depend on rabbitmq_prometheus unconditionally but I wonder if we can do the same thing here and only use the prometheus dependency directly. The downside of depending on rabbitmq_prometheus is that the plugin is enabled as a dependency of this plugin.

What do you think @lukebakken ?

sdewhitt: to clarify, are you asking if you had rabbitmq-stream-s3 depend on prometheus? or are you asking to have rabbitmq-aws use rabbitmq_prometheus directly?

lukebakken: Yeah he's suggesting that this plugin depend on `rabbitmq_prometheus` the [same way that `rabbitmq-stream-s3` does](https://github.com/amazon-mq/rabbitmq-stream-s3/blob/main/Makefile#L5).

I don't see a problem with that.

src/aws_auth_validate_metrics.erl
```diff
@@ -0,0 +1,373 @@
...
+find_bucket_pos(Ms) when Ms =< 10 -> 1;
+find_bucket_pos(Ms) when Ms =< 25 -> 2;
+find_bucket_pos(Ms) when Ms =< 50 -> 3;
+find_bucket_pos(Ms) when Ms =< 100 -> 4;
+find_bucket_pos(Ms) when Ms =< 250 -> 5;
+find_bucket_pos(Ms) when Ms =< 500 -> 6;
+find_bucket_pos(Ms) when Ms =< 1000 -> 7;
+find_bucket_pos(Ms) when Ms =< 2500 -> 8;
+find_bucket_pos(Ms) when Ms =< 5000 -> 9;
+find_bucket_pos(_Ms) -> 10.
```
lukebakken: **Histogram is malformed: no `+Inf` bucket, and slow durations are misattributed.** This fall-through clause puts any duration > 5000ms (including anything > 10000ms) into slot 10, i.e. the `le="10000"` bucket, so a 30s validation is reported as `<=10s`. Combined with `build_cumulative_buckets/2` emitting only the 10 finite boundaries, the histogram has no `le="+Inf"` bucket at all. I confirmed `prometheus_model_helpers:histogram_buckets/1` does not append one. Prometheus requires a `+Inf` bucket equal to `sample_count`; without it `histogram_quantile()` returns NaN/garbage and any SLO alert keyed on the 10s bucket silently misfires.

Suggest adding an overflow slot (an 11th bucket for `Ms > 10000`) and emitting a trailing `{infinity, sample_count}` bucket.

src/aws_auth_validate_metrics.erl
```diff
@@ -0,0 +1,373 @@
...
+            emit_capacity_exhausted_total(Counters, Callback),
+            emit_semaphore_gauges(Callback),
+            ok
+    end.
+
+%%--------------------------------------------------------------------
+%% Internal: observation helpers
+%%--------------------------------------------------------------------
+
+observe_request(Counters, Method, Category) ->
```
lukebakken, line 174: **`unknown_method` requests can never be recorded.** For an unrecognized method path the category is `unknown_method`, but the method binding (e.g. `<<"bogus">>`) is not one of ldap/http/oauth/tls, so `method_index/1` -> `undefined`, `request_slot/2` -> `undefined`, and this clause skips the observation. Category index 12 in `?CATEGORIES` is therefore dead, and probes/scans against bad method names never appear in `requests_total` under any label. Consider a synthetic `method="unknown"` (or an unlabeled) slot so these are still counted.

src/aws_auth_validate_metrics.erl
```diff
@@ -0,0 +1,373 @@
...
+    Buckets.
+
+%% @doc Read current semaphore usage without crashing. Returns
+%% {InUse, Capacity} or `unavailable` if the process is not running.
+try_semaphore_usage() ->
+    try
+        aws_auth_validate_semaphore:usage()
+    catch
+        exit:{noproc, _} -> unavailable;
+        exit:{timeout, _} -> unavailable
```
lukebakken: **A non-`noproc`/`timeout` exit here crashes the whole scrape.** `try_semaphore_usage/0` catches only `exit:{noproc,_}` and `exit:{timeout,_}`. If the semaphore worker exits for any other reason (`shutdown`, `killed`, `{nodedown,_}`, ...), the exception escapes `collect_mf/2` and fails the scrape for **every** collector on `/api/metrics`, not just this one. Given the comment intent is to be defensive about the metrics endpoint, widen this to a catch-all (e.g. `exit:_ -> unavailable`, or `_:_`) that degrades to `unavailable`.

src/aws_auth_validate_mgmt.erl
```diff
@@ -317,7 +317,10 @@ audit(Method, SourceIP, User, ResultCategory, T0) ->
     ?AWS_LOG_INFO(
         "auth_validate: method=~ts user=~ts source_ip=~ts result=~ts duration_ms=~B",
         [sanitize_log(Method), sanitize_log(User), format_ip(SourceIP), ResultCategory, Duration]
-    ).
+    ),
+    %% Emit Prometheus metrics for this request. observe/3 is a no-op when
+    %% the metrics collector is not registered, so this is always safe.
+    aws_auth_validate_metrics:observe(Method, ResultCategory, Duration).
```
lukebakken, line 323: **Fast-reject paths contaminate the latency histogram.** `audit/5` unconditionally calls `observe/3` for every terminal result, and `observe_duration/3` is gated only by `method_index`, not by category. So `body_too_large`, `input_invalid`, and `capacity_exhausted` on a known method feed the low buckets with near-zero durations. Since the histogram is labeled by method only (not result), operators can't separate real LDAP/HTTP round-trip latency from rejection latency, skewing p50/p99 downward. Consider observing duration only for categories that actually ran a backend validation.

src/aws_auth_validate_metrics.erl
```diff
@@ -0,0 +1,373 @@
...
+%% persistent_term keys
+-define(PT_COUNTERS, {?MODULE, counters}).
+
+%%--------------------------------------------------------------------
+%% Registration
+%%--------------------------------------------------------------------
+
+-spec register() -> ok.
+register() ->
+    Counters = counters:new(?TOTAL_SLOTS, [write_concurrency]),
```
lukebakken: **`register/0` overwrites any existing counters ref.** It unconditionally allocates a fresh `counters:new/2` and `persistent_term:put`s it. If `register/0` is ever invoked a second time while metrics are already active (e.g. an `aws_sup` restart re-running `auth_validation_children/0`), all `requests_total`/histogram counts reset to zero mid-interval, producing a counter reset that corrupts `rate()`. I confirmed the overwrite; I did not trace a concrete double-invocation path, so this is conditional, but guarding with a get-or-create (only `put` when absent) would remove the hazard.

src/aws_auth_validate_metrics.erl
```diff
@@ -0,0 +1,373 @@
...
+find_bucket_pos(_Ms) -> 10.
+
+%% @doc Build cumulative bucket list from per-bucket counters.
+%% Prometheus histograms require cumulative counts: each bucket's count
+%% includes all observations in lower buckets.
+build_cumulative_buckets(Counters, Base) ->
+    Boundaries = ?HISTOGRAM_BUCKETS,
+    {Buckets, _} = lists:mapfoldl(
+        fun(Boundary, CumAcc) ->
+            Pos = find_bucket_pos(Boundary),
```
lukebakken: Cleanup (non-blocking): `build_cumulative_buckets/2` recovers each slot index by calling `find_bucket_pos(Boundary)` on a boundary whose slot is already known by its position in `?HISTOGRAM_BUCKETS`. This couples emission correctness to `find_bucket_pos/1`'s boundary equalities; if a boundary were ever duplicated or the lists drift, slots misalign silently. A `lists:zip` of boundaries with slots `1..?NUM_BUCKETS` (or an enumerated fold) expresses the same layout directly.

lukebakken (changes requested):
Thanks for adding a pull-based Prometheus collector for the auth-validation endpoint. The slot layout and the no-op-when-unregistered guard are clean. A few issues need addressing before merge, most severity in the duration histogram and scrape robustness. Details inline; findings 1-3 are the blockers, 4 is a real metric-quality bug, 5-6 are lower priority.

Verified locally at deps/aws on feature/auth-validation-metrics (HEAD c25c10e), including that prometheus_model_helpers:histogram_buckets/1 maps buckets 1:1 with no auto-appended +Inf.

src/aws_auth_validate_metrics.erl
sdewhitt: Fixed in 097938e. Added an 11th overflow bucket for \`Ms > 10000\` and \`build_cumulative_buckets/2\` now emits an explicit trailing \`{infinity, SampleCount}\` so \`le="+Inf"\` == sample_count. CT case \`inf_bucket_exists\` covers it.

src/aws_auth_validate_metrics.erl
sdewhitt: Fixed in 097938e. Added a synthetic \`method="unknown"\` slot so unrecognized-method requests are counted in \`requests_total\` (fixed cardinality -- the raw path segment is never used as a label). CT case \`unknown_method_counted\`.

src/aws_auth_validate_metrics.erl
sdewhitt: Fixed in 097938e. \`try_semaphore_usage/0\` now has a final \`_:_ -> unavailable\` catch-all, so any semaphore exit degrades that gauge to absent instead of failing the whole scrape. CT case \`semaphore_exit_catch_all\`.

src/aws_auth_validate_mgmt.erl
sdewhitt: Fixed in 097938e. Duration is now gated by \`is_timed_category/1\` -- only categories that attempted backend/outbound work are recorded. \`query_invalid\` is excluded because query parsing happens in \`parse_input/1\` before any connection. \`requests_total\` still counts all categories. CT case \`duration_gating_excludes_fast_rejects\`.

src/aws_auth_validate_metrics.erl
sdewhitt: Fixed in 097938e. \`register/0\` is now get-or-create: it only allocates + \`persistent_term:put\`s the counters when absent, and always re-registers the collector, so a re-invocation preserves existing counts. CT case \`register_preserves_counters\`.

src/aws_auth_validate_metrics.erl
sdewhitt: Done in 097938e. \`build_cumulative_buckets/2\` now enumerates boundaries against slots \`1..NUM_BUCKETS\` directly instead of recovering the index via \`find_bucket_pos/1\`.

lukebakken (approved):
All six findings from the earlier review are addressed in 097938e, each with test coverage, and the prometheus dependency now goes through rabbitmq_prometheus per the Makefile thread. Approving.

- +Inf/overflow bucket added (position 11); build_cumulative_buckets/2 emits {infinity, sample_count}, which renders as le="+Inf"
- unknown methods counted under a fixed method="unknown" label (no raw-string cardinality risk)
- try_semaphore_usage/0 has a _:_ catch-all so an arbitrary semaphore exit no longer fails the whole scrape
- duration histogram gated by is_timed_category/1 so pre-connection rejects no longer skew latency
- register/0 is get-or-create, preserving counts across re-registration
- build_cumulative_buckets/2 enumerates slots directly instead of round-tripping through find_bucket_pos/1

Verified locally at deps/aws on feature/auth-validation-metrics (HEAD 097938e): compile, dialyzer (after a PLT rebuild, since the dependency set changed), eunit (183), and ct-aws_auth_validate_metrics (13 Ok / 0 failed) all pass.

lukebakken:
Ah OK. Well, I don't think it's too much of a hassle / hardship since `rabbitmq_prometheus` is tier-1.

PR #185, opened 2026-08-06
Include referenced ARNs in auth-validation audit log

sdewhitt:
Includes the ARN references a validation request carried in its audit-log line.

An ARN is an identifier, not a secret, so logging it is safe and gives operators an attributable record of which secret/certificate a given validation attempt referenced. The resolved secret content itself is never logged.

- New `arns=` field lists the ARN references found in a bounded key set (`password_arn`, `cacertfile_arn`, `certfile_arn`, `keyfile_arn`), including one level into `ssl_options`; emits `arns=none` when there are none.
- Each value is sanitized before interpolation to prevent log injection.
- Inline tokens are never treated as ARNs and never logged.

### Tests
Top-level and nested extraction, `arns=none`, control-character sanitization, token exclusion, and a no-secret-leakage check. Full suite green.

### DIFF
--- a/src/aws_auth_validate_mgmt.erl
+++ b/src/aws_auth_validate_mgmt.erl
@@ -32,8 +32,18 @@
 %% sanitize_log/1 strips control chars from a caller-influenceable audit field;
 %% username/1 extracts the authenticated principal (or the `unknown' fallback);
 %% referenced_arns/2 extracts known ARN reference strings from the request body
-%% for audit logging (never resolved content, only identifiers).
--export([status_for_category/1, max_body_size/0, sanitize_log/1, username/1, referenced_arns/2]).
+%% for audit logging (never resolved content, only identifiers);
+%% encode_arn_value/1 percent-encodes ambiguous bytes for the audit format;
+%% prepare_arn_value/1 validates, caps, and encodes a single ARN value for logging.
+-export([
+    status_for_category/1,
+    max_body_size/0,
+    sanitize_log/1,
+    username/1,
+    referenced_arns/2,
+    encode_arn_value/1,
+    prepare_arn_value/1
+]).
 -endif.
 
 -include_lib("rabbitmq_web_dispatch/include/rabbitmq_web_dispatch_records.hrl").
@@ -42,6 +52,11 @@
 
 -define(DEFAULT_MAX_BODY_SIZE, 65_536).
 -define(DEFAULT_REQUIRED_USER_TAG, administrator).
+%% Maximum byte length of a single ARN value in the audit log. Real ARNs are
+%% well under this limit; the cap prevents a maliciously large value from
+%% producing an unbounded log line. Values exceeding this are truncated with
+%% a "...[truncated]" marker.
+-define(MAX_ARN_LOG_LEN, 2048).
 
 dispatcher() -> [{"/aws/auth/validate/:method", ?MODULE, []}].
 
@@ -356,6 +371,43 @@ sanitize_byte(B) when B < 16#20; B =:= 16#7F ->
 sanitize_byte(B) ->
     <<B>>.
 
+%% Percent-encode bytes that would be ambiguous in the space/comma-delimited
+%% audit format: space (0x20), equals (0x3D), comma (0x2C), percent (0x25),
+%% and control bytes (< 0x20, 0x7F). The encoding is reversible and greppable.
+encode_arn_value(Bin) when is_binary(Bin) ->
+    <<<<(encode_arn_byte(B))/binary>> || <<B>> <= Bin>>.
+
+encode_arn_byte(B) when B < 16#20 -> percent_hex(B);
+encode_arn_byte(16#20) -> <<"%20">>;
+encode_arn_byte(16#25) -> <<"%25">>;
+encode_arn_byte(16#2C) -> <<"%2C">>;
+encode_arn_byte(16#3D) -> <<"%3D">>;
+encode_arn_byte(16#7F) -> <<"%7F">>;
+encode_arn_byte(B) -> <<B>>.
+
+percent_hex(B) ->
+    <<$%, (hex_digit(B bsr 4)), (hex_digit(B band 16#0F))>>.
+
+hex_digit(N) when N < 10 -> $0 + N;
+hex_digit(N) -> $A + N - 10.
+
+%% Prepare an ARN value for audit logging: validate shape, cap length,
+%% percent-encode ambiguous bytes. A value exceeding MAX_ARN_LOG_LEN is
+%% truncated with a marker. A value not starting with "arn:" is wrapped
+%% in [non-arn:...] so it cannot masquerade as a real ARN identifier.
+prepare_arn_value(Bin) when byte_size(Bin) > ?MAX_ARN_LOG_LEN ->
+    Truncated = binary:part(Bin, 0, ?MAX_ARN_LOG_LEN),
+    <<(encode_arn_value(Truncated))/binary, "...[truncated]">>;
+prepare_arn_value(<<"arn:", _/binary>> = Bin) ->
+    encode_arn_value(Bin);
+prepare_arn_value(Bin) ->
+    Prefix =
+        case byte_size(Bin) > 128 of
+            true -> binary:part(Bin, 0, 128);
+            false -> Bin
+        end,
+    <<"[non-arn:", (encode_arn_value(Prefix))/binary, "]">>.
+
 %% The authenticated management user's name, for the audit trail. This endpoint
 %% is admin-gated and makes outbound connections to caller-chosen targets (an
 %% SSRF surface), so the acting principal -- not just the TCP peer IP, which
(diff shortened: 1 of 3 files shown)


src/aws_auth_validate_mgmt.erl
```diff
@@ -312,11 +315,23 @@ decode_json(<<>>) ->
...
+%% lists the ARN reference strings the request carried (identifiers, not resolved
+%% content -- safe to log per the R6 no-credential-leakage invariant). For
+%% pre-decode failures (body_too_large, malformed JSON) arns=none is emitted since
+%% no body was parsed.
+audit(Method, SourceIP, User, ResultCategory, Arns, T0) ->
     Duration = erlang:monotonic_time(millisecond) - T0,
     ?AWS_LOG_INFO(
-        "auth_validate: method=~ts user=~ts source_ip=~ts result=~ts duration_ms=~B",
-        [sanitize_log(Method), sanitize_log(User), format_ip(SourceIP), ResultCategory, Duration]
+        "auth_validate: method=~ts user=~ts source_ip=~ts result=~ts duration_ms=~B arns=~ts",
```
lukebakken, line 341: Log injection: `arns=` is the last field of this space-delimited line, and `sanitize_log` (sanitize_byte at line 351) only replaces bytes < 0x20 and 0x7F. Spaces and `=` survive, so a caller-controlled ARN value such as `arn:aws:x result=success user=root` injects forged `key=value` pairs. A parser that takes the last `result=`/`user=` then records a forged success/principal, which defeats the attribution this record exists to provide.

Consider quoting or percent-encoding the value (or at minimum escaping space and `=`) so a value can't be mistaken for additional fields.

src/aws_auth_validate_mgmt.erl
```diff
@@ -366,6 +381,64 @@ format_ip(IP) when is_tuple(IP) ->
...
+    TlsArnKeys = [<<"cacertfile_arn">>, <<"certfile_arn">>, <<"keyfile_arn">>],
+    case aws_auth_validate_registry:lookup_backend(Method) of
+        {error, unknown_method} ->
+            <<"none">>;
+        {ok, Module} ->
+            AllowedFields = aws_auth_validate_registry:effective_allowed_fields(Module, Method),
+            %% Collect top-level ARN values that are in both the known set and
+            %% the effective allowed fields for this backend.
+            TopLevelArns = [
+                sanitize_log(V)
```
lukebakken: The value is logged verbatim with no length bound and no ARN-format validation, despite being framed as an "ARN reference (identifier)". It is really arbitrary allowed-field content up to `max_body_size` (64 KiB default, configurable to 1 MiB), so a ~60 KiB `password_arn` produces a single ~60 KiB audit line of attacker-chosen non-ARN text (log flooding + a mislabeled field). Suggest bounding the logged length and, ideally, validating the `arn:` shape before treating it as an identifier.

src/aws_auth_validate_mgmt.erl
```diff
@@ -366,6 +381,64 @@ format_ip(IP) when is_tuple(IP) ->
...
+             || K <- KnownArnKeys,
+                lists:member(K, AllowedFields),
+                V <- [maps:get(K, BodyMap, undefined)],
+                is_binary(V)
```
lukebakken: Guard is `is_binary(V)` rather than `is_binary(V) andalso V =/= <<>>`. An empty-string value (`{"password_arn":""}`) is emitted as an empty entry, so the field becomes `arns=` instead of the documented `arns=none`; two empty keys yield `arns=,` (bare comma). A consumer expecting `arns=none` when no ARN is present will misclassify these.

src/aws_auth_validate_mgmt.erl
```diff
@@ -366,6 +381,64 @@ format_ip(IP) when is_tuple(IP) ->
...
+                                ];
+                            _ ->
+                                []
+                        end;
+                    false ->
+                        []
+                end,
+            case TopLevelArns ++ SslArns of
+                [] -> <<"none">>;
+                Collected -> iolist_to_binary(lists:join(<<",">>, Collected))
```
lukebakken, line 508: Values are comma-joined, but `sanitize_log` does not escape commas inside an individual value, so a comma-bearing value is indistinguishable from the separator: `arn:aws:a,arn:aws:b` in one field reads as two ARNs to anything splitting on `,`. Pairs with the injection concern above; a quoting/encoding scheme would resolve both.

src/aws_auth_validate_mgmt.erl
```diff
@@ -366,6 +381,64 @@ format_ip(IP) when is_tuple(IP) ->
...
+referenced_arns(Method, BodyMap) ->
+    KnownArnKeys = [
+        <<"password_arn">>, <<"cacertfile_arn">>, <<"certfile_arn">>, <<"keyfile_arn">>
+    ],
+    TlsArnKeys = [<<"cacertfile_arn">>, <<"certfile_arn">>, <<"keyfile_arn">>],
+    case aws_auth_validate_registry:lookup_backend(Method) of
+        {error, unknown_method} ->
+            <<"none">>;
+        {ok, Module} ->
+            AllowedFields = aws_auth_validate_registry:effective_allowed_fields(Module, Method),
```
lukebakken, line 472: Minor efficiency: `referenced_arns/2` recomputes `lookup_backend/1` and `effective_allowed_fields/2` (which reads app env and filters lists) per request, and `dispatch/2` does the same work again moments later on the success path. Not a correctness issue, but the resolved module/allowed-fields could be threaded through to avoid doing it twice.

src/aws_auth_validate_mgmt.erl
```diff
@@ -366,6 +381,64 @@ format_ip(IP) when is_tuple(IP) ->
...
+%% Top-level ARN keys are included only if they appear in the backend's
+%% effective_allowed_fields. TLS-material ARN keys nested one level inside
+%% ssl_options are included when ssl_options itself is an allowed field (the
+%% parent's presence is sufficient; sub-keys are not individually listed in
+%% effective_allowed_fields).
+referenced_arns(Method, BodyMap) ->
+    KnownArnKeys = [
+        <<"password_arn">>, <<"cacertfile_arn">>, <<"certfile_arn">>, <<"keyfile_arn">>
+    ],
+    TlsArnKeys = [<<"cacertfile_arn">>, <<"certfile_arn">>, <<"keyfile_arn">>],
```
lukebakken: `TlsArnKeys` includes `certfile_arn` and `keyfile_arn`, but for method `tls` the backend accepts only `cacertfile_arn`; `certfile_arn`/`keyfile_arn` under `ssl_options` are rejected as unknown keys. Such a request is rejected `input_invalid` and those ARNs are never resolved, yet they are still recorded in `arns=`, over-reporting what the request effectively referenced.

lukebakken (changes requested):
Thanks for adding ARN attribution to the audit trail. The mechanism is sound (all seven audit/6 call sites updated, extraction gated by effective_allowed_fields so arbitrary body keys can't leak), but a few issues in how the value is produced and logged undercut the audit-integrity goal. Details are inline; the first two are the ones I'd like addressed before merge, the rest are lower severity.

Verified locally at deps/aws on feature/auth-validation-audit-arns (HEAD 3ade7fb).

src/aws_auth_validate_mgmt.erl
sdewhitt: Fixed in f30fcfc -- ARN values are now percent-encoded before logging (space `%20`, `=` `%3D`, `,` `%2C`, `%` `%25`, control/DEL), so a value can no longer inject forged `key=value` pairs. Reversible and greppable.

src/aws_auth_validate_mgmt.erl
sdewhitt: Fixed in f30fcfc -- values are capped at 2048 bytes with a `...[truncated]` marker, and a value not starting with `arn:` is wrapped as `[non-arn:<first-128B>]` so it cannot masquerade as an ARN identifier.

src/aws_auth_validate_mgmt.erl
sdewhitt: Fixed in f30fcfc -- the extraction now guards `V =/= <<>>`, so empty-string values are treated as absent and `arns=none` is emitted as documented.

src/aws_auth_validate_mgmt.erl
sdewhitt: Fixed in f30fcfc -- resolved by the same percent-encoding as the injection fix: a comma inside a value becomes `%2C`, unambiguous from the `,` join separator.

src/aws_auth_validate_mgmt.erl
sdewhitt: Left as-is with an explanatory note (f30fcfc): threading the resolved module/allowed-fields from `referenced_arns/2` into `dispatch/2` would tangle the mgmt->registry boundary for a marginal, non-hot-path saving. Happy to revisit if you prefer.

src/aws_auth_validate_mgmt.erl
sdewhitt: Fixed in f30fcfc -- `ssl_arn_keys_for_method/1` returns only `cacertfile_arn` for `tls` (all three for ldap/http), so ARNs the method never accepts are no longer recorded.

the-mikedavis (approved):

PR #186, opened 2026-08-06
Precompile auth-validation UI templates at build time

sdewhitt:
Fixes the **Auth Validation** management-console tab, which failed to load with:

```
Error: Template not found: aws_auth_validate
```

### Cause
The management UI changed how it loads EJS templates: instead of fetching
`js/tmpl/<name>.ejs` at runtime, it now looks the template up in a
build-time-generated `COMPILED_TEMPLATES` map (removing the runtime `eval`,
which lets the UI drop the `unsafe-eval`/`unsafe-inline` CSP directives). This
plugin still shipped only the raw `.ejs`, so its template was never compiled
into the map and the lookup threw.

### Fix
Mirror what every other management-UI plugin already does:

- Add `rabbitmq-management-plugin.mk` to `DEP_PLUGINS` so `make app`
  precompiles `priv/www/js/tmpl/*.ejs` into `priv/www/js/aws-ejs.js`.
- Register `aws-ejs.js` ahead of the view's JS in `web_ui/0` so the compiled
  templates are present before the view renders them.
- Gitignore the generated bundle.

### Verification
`make app` regenerates `aws-ejs.js` containing the compiled
`aws_auth_validate` template; the build's EJS sanity check passes. The raw
`.ejs` remains the source of truth.


Makefile
```diff
@@ -15,7 +15,10 @@ LOCAL_DEPS = crypto inets ssl xmerl public_key eldap
 PLT_APPS = rabbit
 
 DEP_EARLY_PLUGINS = rabbit_common/mk/rabbitmq-early-plugin.mk
-DEP_PLUGINS = rabbit_common/mk/rabbitmq-plugin.mk
+# Precompile the auth-validation UI's EJS templates into priv/www/js/aws-ejs.js
+# at build time (the management UI no longer compiles .ejs at runtime).
+DEP_PLUGINS = rabbit_common/mk/rabbitmq-plugin.mk \
+              rabbitmq_management/mk/rabbitmq-management-plugin.mk
```
the-mikedavis: How portable is this between versions? I think this was only added upstream recently if I remember correctly (?), so it might not work for 4.2 or 3.13

sdewhitt: It is not portable, investigating doing a version gate approach

the-mikedavis (approved):

PR #188, opened 2026-08-20
Add node-health detector for attributing a degraded cluster node

lukebakken:
> [!NOTE]
> This PR description was written by Claude (Anthropic's Claude Code) under the direction of @lukebakken, who reviewed it before opening.

## Summary

Adds an off-by-default detector that attributes a partially-degraded cluster node -- one whose host or uplink drops a large fraction of packets while the node still answers and participates, so reachability-based checks miss it -- from the per-peer reachability estimates the node failure detector already produces. Each node samples its own view, gossips it to peers so every node holds the full observer-by-peer matrix, and computes a verdict that is exposed as Prometheus gauges a simple alarm can act on.

## What's here

- `aws_node_health` -- pure scorer. Given a rolling window of observer-by-peer probability snapshots it returns `clean` | `{suspect, Node}` | `cluster_wide`. Three attribution paths plus a guard, evaluated **P2 -> P3 -> cluster-wide guard -> P1**:
  - **P2 (sustained)**: a node extreme for most of the window and dominating the next by a margin (continuous fault).
  - **P3 (bidirectional)**: a node elevated inbound whose own outbound row shows every peer degraded *and* dominates the others' by a margin -- attributes a single fault masked by cluster-wide congestion, which an inbound-only threshold cannot separate from a congestion-elevated healthy node.
  - **cluster-wide guard**: two or more nodes elevated with no dominant culprit -> `cluster_wide`, not blamed on any node.
  - **P1 (isolated, flap-rate)**: the most-flapping node (by extreme-threshold crossings, not by median -- a low-duty-cycle fault has a near-zero median), when every other node is pristine on both median and flap count. This is the intermittent/periodic path, where a fraction-of-time measure starves on the low duty cycle.
  - Attribution runs only with three or more observed nodes; below that (a transient rolling-restart state) the verdict is `clean`.
- `aws_node_health_worker` -- gossip gen_server: samples locally (bounded), casts rows to peers, evicts stale rows (from both the per-tick snapshot and its persistent state, so replaced node names do not accumulate), keeps the bounded window, recomputes the verdict, and **debounces** the published `suspected` flag and the `cluster_wide` signal with `confirm_ticks`/`clear_ticks` hysteresis (a confirmed single-node suspect takes precedence over `cluster_wide`, so the two never both fire). The cycle path is hardened against the degraded-cluster conditions it observes: the sampling call is time-bounded, the metadata-store peer-list read is guarded, and gossiped rows with non-numeric values are rejected before reaching the scorer.
- `aws_node_health_sup` -- dedicated sub-supervisor for the worker, so a crash-looping worker consumes only its own restart budget and cannot cascade to the unrelated auth-validation feature.
- `aws_node_health_metrics` -- `prometheus_collector` emitting, at scrape time and crash-safe:
  - `rabbitmq_aws_node_health_peer_down_probability{peer}` -- this node's own raw estimate (no gossip needed).
  - `rabbitmq_aws_node_health_peer_down_suspected{peer}` -- `1` if `peer` is attributed as the single degraded node.
  - `rabbitmq_aws_node_health_peer_down_confidence{peer}` -- confidence in `[0,1]`.
  - `rabbitmq_aws_node_health_cluster_congested` -- unlabelled `1` when degradation is symmetric across the cluster.
- `aws_node_health_config` -- single home for the feature's environment reads, defaults, and bounds.
- `aws_sup` wiring behind `aws.node_health.enabled` (default off), plus the `aws.node_health.*` schema mappings (with round-trip config-schema coverage for every key).

## Consumer model

The intended consumer is a **dumb CloudWatch alarm**, which cannot correlate across nodes the way PromQL can. That is why the verdict is computed in-broker (where the gossiped matrix is assemblable) and exposed as a ready-to-threshold `suspected` / `cluster_congested` signal, rather than exporting only raw per-node rows for downstream correlation. The metric names use the plugin's `rabbitmq_aws_<feature>_*` convention; the underlying failure detector (`aten`) is an implementation detail and is never named in the metrics.

## Testing

- **eunit** -- the pure scorer (all paths + hysteresis against captured probability fixtures, including regressions for the low-duty-cycle P1 candidate, P3 requiring the other nodes' rows to be present, and the sub-three-node guard), the worker (including row validation and stale-row pruning), the collector sample builders, and config.
- **Common Test** (`aws_node_health_SUITE`) -- boots a real broker, lets `aws_sup` start the worker and register the collector from config, and verifies the integration layer: the toggle, the live collector rendering exactly the `rabbitmq_aws_node_health_*` families through the real prometheus text format, and the verdict paths (clean / `{suspect,_}` via P2 / `cluster_wide`) driven by injecting synthetic peer rows through the worker's public API.
- **In-vivo** -- validated on a 3-node cluster under `tc/netem`: periodic loss (P1), continuous loss (P2, probability stays pinned high -- it does not re-normalise), masked single-node fault under uniform-loss background (P3), and symmetric congestion (`cluster_congested`, driven with latency+jitter to avoid tripping `net_tick`). No healthy node was ever misattributed.
- Gates: `eqwalizer` clean, `erlfmt`, `dialyze`, and the config-schema suite all pass.

## Known limitations

- The signal measures Erlang-distribution reachability, so it identifies the degraded **node**, not the cause -- a GC/CPU stall looks like a network fault (documented in `NODE_HEALTH.md`).
- Attribution needs >= 3 observed nodes (Amazon MQ clusters are three-node); below that it returns `clean` rather than guess, and being pre-partition it reverts once a node is fully partitioned out.
- Scorer thresholds are hardcoded defaults calibrated on a 3-node topology; only the operational knobs (`window`, `interval_ms`, `stale_ticks`, `confirm_ticks`, `clear_ticks`) are configurable.
- Under sustained *symmetric* congestion, a structurally-asymmetric node (more queue leaders / a load-balancer target) can read enough of an inbound lead to transiently trip P2's margin and be attributed as a single-node suspect instead of reporting `cluster_wide` -- a false positive on a healthy node (never a missed fault), intermittent and boundary-dependent, observed in vivo under uniform latency+jitter. P2 has no own-outbound dominance check (P3 does), and bidirectionality does not screen it out because uniform congestion makes every node bidirectional. Tracked in #191.

## Related / follow-ups

- #190 -- deregister Prometheus collectors on feature disable / app stop (plugin-wide, not specific to this feature).
- #191 -- P2 can false-attribute the busiest node under symmetric congestion; proposed fix is an own-outbound dominance gate on P2 (with a missing-row fallback).
- Consumer-side alarm wiring is out of scope for this PR.


the-mikedavis (approved):
I have this a quick read through and it looks good to me, and seems extra low-risk since it's off-by-default and is observational-only

